PR #1, opened 2015-04-02
Fix NULL + Improved BQSQLException

jsiessataccess:
Hi Jonathan,

Thank you for sharing your fork.

The first commit in this pull request fixes null handling for float (and probably other datatypes). This fixes https://code.google.com/p/starschema-bigquery-jdbc/issues/detail?id=13

1) The issue was to use 
result == null
instead of 
com.google.api.client.util.Data.isNull(result) a.k.a. Data.isNull(result)
as documented in TableCell.getV()

The second commit is an improvement in the error message.

2) Most of the valuable information is in the detail message of the GoogleJsonResponseException, so I think it is better to append the detail message to the main message.
I also fixed two exceptions from SQLException to BQSQLException to be consistent accross the driver (raising only BQSQLException).

3) Third commit fixes Issue 10: BigQuery Integer is bigger than a Java Integer: https://code.google.com/p/starschema-bigquery-jdbc/issues/detail?id=10

Sincerely,

jsiess


jonathanswenson:
@jsiessataccess Thanks so much for your work on this. 

I'm slowly getting around to pulling in pieces of this PR. I should be merging in the first several commits in the near future.

jonathanswenson:
@jsiessataccess I have merged in several of the commits in this PR. 

In the future it would be preferable for each PR to contain a single bugfix or feature. In order to merge the remaining commits, please open a PR for each distinct unit of work.

PR #11, opened 2015-11-04
Auth enhance

stalbot:
Changes so the the user can pass in a password via the JDBC url by passing in a path. Old functionality still works (no path, password is assumed to be a path to a p12 file with password 'notasecret'.).

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -104,6 +104,7 @@ public class BQConnection implements Connection {
         boolean containUserPassword = false;
         String userId;
         String userKey;
+        String userPath;
 
         boolean serviceAccount = false;
 
@@ -122,14 +123,11 @@ public class BQConnection implements Connection {
         //getting the user/password for the connection
         if (containUserPassword) {
             //getting the User/Password from the URL
-            int passwordindex = url.indexOf("&password=");
-            int userindex = url.indexOf("&user=");
             try {
-                userId = URLDecoder.decode(url.substring(
-                        userindex + "&user=".length(), passwordindex), "UTF-8");
-                userKey = URLDecoder.decode(
-                        url.substring(passwordindex + "&password=".length()),
-                        "UTF-8");
+                Map<String, String> components = BQSupportFuncts.getUrlQueryComponents(url);
+                userId = components.get("user");
+                userKey = components.get("password");
+                userPath = components.get("path");
             } catch (UnsupportedEncodingException e2) {
                 throw new BQSQLException(e2);
             }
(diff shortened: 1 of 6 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -122,14 +123,11 @@ public BQConnection(String url, Properties loginProp) throws SQLException {
...
-                userId = URLDecoder.decode(url.substring(
-                        userindex + "&user=".length(), passwordindex), "UTF-8");
-                userKey = URLDecoder.decode(
-                        url.substring(passwordindex + "&password=".length()),
-                        "UTF-8");
+                Map<String, String> components = BQSupportFuncts.getUrlQueryComponents(url);
+                userId = components.get("user");
+                userKey = components.get("password");
+                userPath = components.get("path");
             } catch (UnsupportedEncodingException e2) {
```
jonathanswenson: Much better. Thank you.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -197,7 +196,12 @@ public BQConnection(String url, Properties loginProp) throws SQLException {
         //do we have a serviceaccount to connect with?
         if (serviceAccount) {
             try {
-                this.bigquery = Oauth2Bigquery.authorizeviaservice(userId, userKey);
+                // Support for old behavior, passing no actual password, but passing the path as 'password'
+                if (userPath == null) {
+                    userPath = userKey;
+                    userKey = null;
+                }
+                this.bigquery = Oauth2Bigquery.authorizeviaservice(userId, userPath, userKey);
```
jonathanswenson: Yeah, I can't say I LIKE this, but it does preserve backwards compatibility. 

We probably want to produce a README soon. As things are starting to change and the docs are non-existent.

src/test/java/BQJDBC/QueryResultTest/JdbcUrlTest.java
```diff
@@ -33,6 +36,25 @@ public void canRunQueryWithDefaultDataset() throws SQLException {
 
         // This should not blow up with a "No dataset specified" exception
         stmt.executeQuery("SELECT * FROM orders limit 1");
-    }    
+    }
+
+    @Test
+    public void canConnectWithPasswordProtectedP12File() throws SQLException {
+        String url = "jdbc:BQDriver::disco-parsec-659/looker_test?withServiceAccount=true&user=697117590302-76cr6q3217nck6gks0kf4r151j4d9f8e@developer.gserviceaccount.com&password=123456&path=src%2Ftest%2Fresources%2Fbigquery_credentials_protected.p12";
+        BQConnection bqConn = new BQConnection(url, new Properties());
```
jonathanswenson: These properties should come from a serviceaccount.properties file.

PR #13, opened 2016-06-15
Refactor connection param parsing + add userAgent param

jonathanswenson:
Slightly more reusable approach to parameter parsing. All parsed params
are now fully case insensitive (before it was only url parameters that were
fully case sensitive, while params passed in the Properties were only partially
case insensitive.

While it was implicitly done on a case by case basis before, now all jdbc url
parameters will override anything in the login properties. I don't believe that
this is standard in JDBC drivers (though I could be wrong here), this was the
original behavior and I would like to maintain it for backwards compatibility.

Additionally, this PR adds another JDBC property that can be read from the
JDBC url or the login properties to set the userAgent on all BigQuery Requests
from an authenticated BigQuery instance. If this parameter is not used, no
user agent headers will be added to the request.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,5 +149,5 @@
     </plugins>
     <finalName>${project.artifactId}-${project.version}-thin</finalName>
   </build>
-  <version>1.5.1</version>
+  <version>1.6.0</version>
 </project>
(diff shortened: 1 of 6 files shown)


jonathanswenson:
@stalbot Do you have any thoughts?

jonathanswenson:
Going to bump the minor version. These changes should be backwards compatible.

PR #19, opened 2017-03-30
Bump fetch size for BigQuery to much higher number

stalbot:
Pretty self explanatory. 100 is way too low!


jonathanswenson:
is 5k too high? I'm cool with this though. Maybe we want to make this a breaking release bump though.

PR #21, opened 2017-04-28
Fix find column

myl-google:


lewisdawson:
@jonathanswenson Was there a specific reason this fix has never been merged in? It fixes the exception:

```
No Such column labeled: idnet.starschema.clouddb.jdbc.BQSQLException: No Such column labeled
```

that is thrown when one tries to do a column label lookup with `resultSet.get<type>(<columnName>)`.

fzakaria:
Sorry this has not been looked at it.

Can you rebase your changes and tag me; I can review.

fzakaria:
Thank you for the contribution !

PR #23, opened 2017-08-29
Handle a query cancel more gracefully (not NPE)

stalbot:


src/test/java/BQJDBC/QueryResultTest/CancelTest.java
```diff
@@ -30,7 +32,21 @@ public void setup() throws SQLException, IOException {
         this.bq = new BQConnection(url, new Properties());
     }
 
+    @After
+    public void teardown() throws Throwable {
+        if (this.diedWith != null) {
+            throw this.diedWith;
```
stalbot: I couldn't actually get the tests to fail with this pre-fix, but it's a good thing to be checking for anyway. I think we need a longer-running query to reproduce.

PR #24, opened 2017-09-26
Gustav bqjson - accept .json key for connection in addition to .p12

Dutchgoose:
https://github.com/looker/helltool/issues/30636

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,6 +26,12 @@
     </license>
   </licenses>
   <dependencies>
+    <!-- https://mvnrepository.com/artifact/com.google.auth/google-auth-library-oauth2-http -->
+    <dependency>
+        <groupId>com.google.auth</groupId>
+        <artifactId>google-auth-library-oauth2-http</artifactId>
+        <version>0.8.0</version>
+    </dependency>
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
@@ -40,6 +46,16 @@
       <groupId>com.google.api-client</groupId>
       <artifactId>google-api-client</artifactId>
       <version>1.22.0</version>
+    </dependency>
+    <dependency>
+        <groupId>com.google.apis</groupId>
+        <artifactId>google-api-services-oauth2</artifactId>
+        <version>v1-rev134-1.22.0</version>
+    </dependency>
+    <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bigquery</artifactId>
+        <version>0.24.0-beta</version>
     </dependency>
     <dependency>
       <groupId>org.mortbay.jetty</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -26,12 +26,6 @@
     </license>
   </licenses>
   <dependencies>
-    <!-- https://mvnrepository.com/artifact/com.google.auth/google-auth-library-oauth2-http -->
-    <dependency>
-        <groupId>com.google.auth</groupId>
-        <artifactId>google-auth-library-oauth2-http</artifactId>
-        <version>0.8.0</version>
-    </dependency>
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
@@ -46,16 +40,6 @@
       <groupId>com.google.api-client</groupId>
       <artifactId>google-api-client</artifactId>
       <version>1.22.0</version>
-    </dependency>
-    <dependency>
-        <groupId>com.google.apis</groupId>
-        <artifactId>google-api-services-oauth2</artifactId>
-        <version>v1-rev134-1.22.0</version>
-    </dependency>
-    <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-bigquery</artifactId>
-        <version>0.24.0-beta</version>
     </dependency>
     <dependency>
       <groupId>org.mortbay.jetty</groupId>
--- a/src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
+++ b/src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
@@ -204,57 +204,70 @@ public class Oauth2Bigquery {
     }
 
     /**
-     * This function gives back an Authorized Bigquery Client It uses a service
-     * account, which doesn't need user interaction for connect
+     * This function gives back an built GoogleCredential Ojbect from a p12 keyfile
      *
      * @param serviceaccountemail
      * @param keypath
-     * @return Authorized Bigquery Client via serviceaccount e-mail and keypath
+     * @return Built GoogleCredential via serviceaccount e-mail and keypath
      * @throws GeneralSecurityException
      * @throws IOException
      */
-    public static Bigquery authorizeviaservice(String serviceaccountemail,
-                                               String keypath,
-                                               String password,
+    public static GoogleCredential Create_P12_Credential(String serviceaccountemail,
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -26,6 +26,12 @@
     </license>
   </licenses>
   <dependencies>
+    <!-- https://mvnrepository.com/artifact/com.google.auth/google-auth-library-oauth2-http -->
```
Dutchgoose: remove

src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -27,6 +27,12 @@
 
 package net.starschema.clouddb.cmdlineverification;
 
+//import com.google.auth.oauth2.GoogleCredentials;
```
Dutchgoose: remove

src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -220,6 +229,37 @@ public static Bigquery authorizeviaservice(String serviceaccountemail,
                                                Integer readTimeout) throws GeneralSecurityException, IOException {
         logger.debug("Authorizing with service account.");
 
+        /////////////////// NEW METHOD //////////
```
Dutchgoose: remove

src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -43,13 +43,16 @@
 import com.google.api.services.bigquery.BigqueryRequest;
 import com.google.api.services.bigquery.BigqueryRequestInitializer;
 import com.google.api.services.bigquery.BigqueryScopes;
+import net.starschema.clouddb.jdbc.BQSQLException;
```
Dutchgoose: remove

Dutchgoose:
Want to get the checked in so that I can upload this version of the driver to maven for the helltool side of this change. #32269

src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -219,29 +220,41 @@ public static Bigquery authorizeviaservice(String serviceaccountemail,
...
+            credential = builder.build();
         }
+        // If not .p12 must be .json
         else {
-            PrivateKey pk = getPrivateKeyFromCredentials(keypath, password);
-            builder = builder.setServiceAccountPrivateKey(pk);
+            // For .json load the key via credential.fromStream
+            File jsonKey = new File(keypath);
+            InputStream inputStream = new FileInputStream(jsonKey);
+            credential = credential.fromStream(inputStream, CmdlineUtils.getHttpTransport(), CmdlineUtils.getJsonFactory()).createScoped(scopes);
```
jonathanswenson: What happens here if a malformed JSON file is passed? Does this fail gracefully?

jonathanswenson: Does the JSON file include the service account email? Do we need to worry about setting one? Is the service account email parameter just ignored here?

Dutchgoose:
```
{
  "type": "service_account",
  "project_id": "... PROJECT ID",
  "private_key_id": "... PK ID",
  "private_key": "... PRIVATE KEY",
  "client_email": "... CLIENT EMAIL",
  "client_id": "... CLIENT ID",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://accounts.google.com/o/oauth2/token",
  "auth_provider_x509_cert_url": "... CERT GOES HERE",
  "client_x509_cert_url": "... CERT GOES HERE"
}
``` 
format of a json key with all sensitive info removed.

When you generate they key it has to be attached to an email.

Dutchgoose:
```
projectid=disco-parsec-659
type=service
user= EMAIL FOR SERVICE ACCOUNT 
password= PATH TO THE JSON KEY ( same format as above )
dataset= DATASET 
transformquery=true
```

This is the `.properties` file I am testing with for json

Dutchgoose: Without an email - it will fail with a query  when you run the query but still create a connection. `jdbc.BQSQLException` 
 In helltool looks like this - 

screen shot 2017-10-06 at 12 09 06 pm

With a completely borked json keyfile it will throw an IO exception. Which looks like this in helltool.

screen shot 2017-10-06 at 12 10 19 pm

src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -203,35 +204,27 @@ public static Bigquery authorizeviainstalled(String clientid,
...
                 .setTransport(CmdlineUtils.getHttpTransport())
                 .setJsonFactory(CmdlineUtils.getJsonFactory())
                 .setServiceAccountId(serviceaccountemail)
-                        // e-mail ADDRESS!!!!
-                .setServiceAccountScopes(scopes);
-                        // Currently we only want to access bigquery, but it's possible
-                        // to name more than one service too
+                // e-mail ADDRESS!!!!
+                .setServiceAccountScopes(GenerateScopes());
+        // Currently we only want to access bigquery, but it's possible
```
Dutchgoose: re-indent

Dutchgoose:
@jonathanswenson any other feedback?

src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -240,8 +233,41 @@ public static Bigquery authorizeviaservice(String serviceaccountemail,
...
+    /**
+     * This function gives back an Authorized Bigquery Client It uses a service
+     * account, which doesn't need user interaction for connect
+     *
+     * @param credential
+     * @return Authorized Bigquery Client via serviceaccount credential
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    public static Bigquery authorizeviaservice(GoogleCredential credential,
```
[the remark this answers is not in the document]
jonathanswenson: We might want to not break this contract here. Perhaps authorizeviaservice should take the file (as before) and based on the file type decide which of the two methods to call (instead of making the caller decide and change the interface here)

Dutchgoose:
@jonathanswenson another attempt!

src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -240,8 +232,49 @@ public static Bigquery authorizeviaservice(String serviceaccountemail,
...
+                                               String userAgent,
+                                               Integer connectTimeout,
+                                               Integer readTimeout) throws GeneralSecurityException, IOException {
+        GoogleCredential credential;
+        // Determine which keyfile we are trying to authenticate with.
+        if (Pattern.matches(".*\\.json$", keypath)) {
+            credential = Oauth2Bigquery.createJsonCredential(keypath);
+        } else {
+            credential = Oauth2Bigquery.createP12Credential(serviceaccountemail, keypath, password);
+        }
```
jonathanswenson: Yep this was along the lines of what I was thinking to keep the interface the same.

PR #26, opened 2018-08-30
Better type support

stalbot:
In forward only result set, support DATETIME and NUMERIC types, and better fallback for when we don't know a type


conrizzle:
Good call on the version bump... also, super stoked you're doing this!

marrlooker:
Lol I like how this is on Jonny's personal account

PR #29, opened 2018-09-25
Add some resiliency to the API interactions

stalbot:
1. Bump default timeout to 240s, which according to Google is the timeout on their backend.
1. Make sure the job id shows up the error message we throw when the result fetch fails.
1. Add some retries into our result fetch.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
@@ -42,6 +42,7 @@ import java.sql.SQLException;
  */
 public class BQStatement extends BQStatementRoot implements java.sql.Statement {
 
+    public static final int MAX_IO_FAILURE_RETRIES = 3;
     private Job job;
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -108,24 +109,38 @@ private ResultSet executeQueryHelper(String querySql) throws SQLException {
...
-                        "DONE")) {
+
+                String status;
+                try {
+                    status = BQSupportFuncts.getQueryState(referencedJob,
+                            this.connection.getBigquery(),
+                            this.ProjectId.replace("__", ":").replace("_", "."));
+                } catch (IOException e) {
+                    if (retries++ < MAX_IO_FAILURE_RETRIES) {
+                       continue;
```
conrizzle: - Should we sleep before retrying? Presumably the failure is from the serve being overloaded.
- Some debug logging might be useful.

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -108,24 +109,38 @@ private ResultSet executeQueryHelper(String querySql) throws SQLException {
...
+                    } else {
+                        throw new BQSQLException(
```
conrizzle: Might be DRYer to just reraise the exception.

PR #30, opened 2018-11-09
Don't explode when you get an array or hash

stalbot:
~Obviously the format here could be better. This is taking the underlying data from what I assume is the d-jsonified BigQuery API response and re-JSONifying it. Going to at least glance at how hard it would be try to pull metadata out of the response and give things the right structure, rather than this naive JSONification. But if that looks like it will take a bit, it's still much much better to not explode.~

Fixed the format to return basically what you'd expect

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
@@ -27,6 +27,8 @@
 
 package net.starschema.clouddb.jdbc;
 
+import com.google.api.client.json.JsonGenerator;
+import com.google.api.client.json.jackson.JacksonFactory;
 import com.google.api.client.util.Data;
 import com.google.api.services.bigquery.Bigquery;
 import com.google.api.services.bigquery.model.GetQueryResultsResponse;
@@ -42,6 +44,7 @@ import java.net.URL;
 import java.sql.*;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
+import java.util.List;
 import java.util.Map;
 import java.util.TimeZone;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -109,6 +111,16 @@ public static String constructUrlFromPropertiesFile(Properties properties, boole
             return null;
         }
 
+        String useLegacySql = properties.getProperty("useLegacySql");
```
stalbot: Should probably change this up. This is a very silly way to build a url that I just really quickly copied from the code around it trying to get the test to work.

stalbot: Ah, checking, this is only used in tests, so if it works it's fine.

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -271,7 +273,64 @@ public String getString(int columnIndex) throws SQLException {
...
         }
         this.wasnull = false;
-        return (String) resultObject;
+        if (resultObject instanceof List || resultObject instanceof Map) {
+            Object resultTransformedWithSchema = smartTransformResult(
+                    resultObject,
+                    this.Result.getSchema().getFields().get(columnIndex - 1));
+            resultObject = easyToJson(resultTransformedWithSchema);
+        }
+        return resultObject.toString();
```
stalbot: Note this change. Rather than casting, call `toString()`, which should never explode. If the API starts sending us novel types, much better to give back some nonsense string than explode, and often the string _won't_ be nonsense.

PR #31, opened 2018-11-09
version bump

stalbot:
Forgot the version bump in #30 

heads up @jonathanswenson @conrizzle


conrizzle:
Good catch... I wonder if we could write a test for that.

PR #32, opened 2018-11-27
Colons in dataset name

conrizzle:
Match the known JDBC connection string for better parsing of the projectId in order to allow for colons in the projectId. If the expected connection string format is not matched then fallback to the old way of simply using the last occurence of a colon as the start index of the projectId/dataset path params.

While we're in here, fixup a test which expected HashMap serialization ordering to always be consistent.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -110,7 +110,15 @@ public class BQConnection implements Connection {
         this.isclosed = false;
 
         try {
-            String pathParams = URLDecoder.decode(url.substring(url.lastIndexOf(":") + 1, url.indexOf('?')), "UTF-8");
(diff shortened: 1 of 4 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -110,7 +110,15 @@ public BQConnection(String url, Properties loginProp) throws SQLException {
         this.isclosed = false;
 
         try {
-            String pathParams = URLDecoder.decode(url.substring(url.lastIndexOf(":") + 1, url.indexOf('?')), "UTF-8");
+            Pattern pathParamsMatcher = Pattern.compile("^jdbc:BQDriver::?([^?]*)", Pattern.CASE_INSENSITIVE);
```
stalbot: Not that it would be a huge bottleneck, but I think ideally this `Pattern` would be statically compiled once, rather than inline for absolute best Java practices.

PR #33, opened 2019-02-21
Fix MaxBillingBytes Behavior for BigQuery PDTs

kaibankslooker:
See https://looker.atlassian.net/browse/MOD-90 for context. 

Add an `unlimitedBillingBytes` flag to `execute()` and `executeQuery()` so that we can specify in Helltool that BigQuery PDTs should bypass the Max Billing Gigabytes setting. Also includes tests. 

Also... fixes https://github.com/jonathanswenson/starschema-bigquery-jdbc/issues/22

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
@@ -164,7 +164,29 @@ public abstract class BQStatementRoot {
         if (this.isClosed()) {
             throw new BQSQLException("This Statement is Closed");
         }
-        this.resset = this.executeQuery(arg0);
+        this.resset = this.executeQuery(arg0, false);
+        this.logger.info("Executing Query: " + arg0);
+        if (this.resset != null) {
+            return true;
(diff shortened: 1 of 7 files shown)


kaibankslooker:
Note: this does not directly `fix` the JIRA issue. Once this gets merged, I'll need to:

1. build `starschema` and upload the new version to nexus
2. update `helltool-java-gem` to reflect new version
3. update `helltool` with new htjg version and change the call to `execute` in the bigquery file

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
```diff
@@ -164,7 +164,29 @@ public boolean execute(String arg0) throws SQLException {
         if (this.isClosed()) {
             throw new BQSQLException("This Statement is Closed");
         }
-        this.resset = this.executeQuery(arg0);
+        this.resset = this.executeQuery(arg0, false);
```
stalbot: as discussed offline, you may want to have this single-arg execute simply call the double-arg execute

PR #35, opened 2019-10-14
Add metadata support for dates and records

kaibankslooker:
As title suggests: add support for getting info about `DATE` and `RECORD`/`STRUCT`-type columns. 

Before this: `ResultSet.getMetaData().getcolumnType([index of a DATE-type column])` -> `Unsupported Type: DATE`

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQPreparedStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQPreparedStatement.java
@@ -307,6 +307,11 @@ public class BQPreparedStatement extends BQStatementRoot implements
     }
 
     @Override
+    public ResultSet executeQuery(String querySql) throws SQLException {
+        throw new BQSQLFeatureNotSupportedException("executeQuery(String querySQL)");
+    }
+
+    @Override
     public ParameterMetaData getParameterMetaData() throws SQLException {
         logger.debug("function call: getParameterMetaData()");
         // TODO IMPLEMENT
--- a/src/test/java/BQJDBC/QueryResultTest/PreparedStatementTests.java
+++ b/src/test/java/BQJDBC/QueryResultTest/PreparedStatementTests.java
@@ -149,24 +149,44 @@ public class PreparedStatementTests {
     }
 
     /**
-     * This test ensures that getColumnType supports the Date type
+     * This test ensures that getColumnType supports the following types: DOUBLE, BOOLEAN, BIGINT, VARCHAR, TIMESTAMP, DATE
+     * Note: RECORD/STRUCT are not tested here because PreparedStatementsTests run queries on Legacy SQL
      */
     @Test
-    public void ResultSetMetadataFunctionTestDateType() {
-        final String sql = "SELECT CAST(CURRENT_DATE() AS DATE)";
+    public void ResultSetMetadataFunctionTestTypes() {
+        final String[] queries = new String[]{
+                "SELECT 3.14",
+                "SELECT TRUE",
+                "SELECT 1",
+                "SELECT 'test'",
+                "SELECT CAST(CURRENT_TIMESTAMP() AS TIMESTAMP)",
+                "SELECT CAST(CURRENT_DATE() AS DATE)",
+        };
+        final int[] expectedType = new int[]{
+                java.sql.Types.DOUBLE,
+                java.sql.Types.BOOLEAN,
+                java.sql.Types.BIGINT,
+                java.sql.Types.VARCHAR,
+                java.sql.Types.TIMESTAMP,
+                java.sql.Types.DATE,
+        };
 
-        System.out.println("Running query:" + sql);
-        try {
-            PreparedStatement stm = PreparedStatementTests.con
-                    .prepareStatement(sql);
-            java.sql.ResultSet theResult = stm.executeQuery();
-            Assert.assertNotNull(theResult);
-            Assert.assertEquals("DATE type was not returned in metadata",
-                    java.sql.Types.DATE,
-                    theResult.getMetaData().getColumnType(1));
-        } catch (SQLException e) {
-            Assert.fail(e.toString());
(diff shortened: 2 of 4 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQPreparedStatement.java
```diff
@@ -306,6 +306,11 @@ public int executeUpdate() throws SQLException {
         throw new BQSQLFeatureNotSupportedException("executeUpdate()");
     }
 
+    @Override
+    public ResultSet executeQuery(String querySql) throws SQLException {
+        throw new BQSQLFeatureNotSupportedException("executeQuery(String querySQL)");
+    }
+
```
jonathanswenson: I take it this is there to implement the Statement Interface? By doing this, the default way to run queries with a string (part of the JDBC spec) would be unsupported. I'm not sure if anyone actually uses this with prepared statements. 

It looks like this problem is originally from a change that was made to 
https://github.com/jonathanswenson/starschema-bigquery-jdbc/blob/275a459c56058191206bcaced99f0799e893f1ca/src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java#L243

Here we added the functionality to pass in unlimitedBillingBytes. So what probably happened is that we had a cached version of BQPreparedStatement.class around so it didn't recompile because it didn't realize it had to when the base class changed. 

Would it make sense to just add an executeQuery method to BQStatementRoot that has a default value (such as unlimitedBillingBytes set to false instead)?

jonathanswenson: Ah, yeah it looks like we just tweaked the basic implementation in the BQStatement class in https://github.com/jonathanswenson/starschema-bigquery-jdbc/pull/33/files

kaibankslooker: I was toying with the idea of adding a default-valued `executeQuery(String querySql)` to the root statement - however, there is a comment in the base definition of `Statement.java` for `executeQuery(String querySql)` that says "Note: This method cannot be called on a PreparedStatement or CallableStatement." 

I think it's `executeQuery()` (no parameters) that's getting called for PreparedStatements - is there actually a use case for `executeQuery(String querySql)` that I'm missing?

kaibankslooker: After all, a prepared statement already has its SQL defined, so calling for the execution of a query with different SQL is a bit weird

kaibankslooker: Ah but I see what you mean, it's a bit weird to override a function that shouldn't exist on this interface - it should really exist on the base Statement class. I'll make that change.

jonathanswenson: Technically it MUST be there for both. That's part of the statement interface even though it doesn't really make sense...

jonathanswenson: https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#executeQuery(java.lang.String)

jonathanswenson: Though because of what you said, it might be ok to just have an unimplemented error. 

What was the behavior before we added the billingbytes flag?

kaibankslooker: i.e. calling `PreparedStatement.executeQuery(sql)` pre-billingbytes change? Hmm, I can try to figure that out. Also note that even the docs you sent indicate that "Note:This method cannot be called on a PreparedStatement or CallableStatement." so maybe it does make sense to keep this unimplemented? If someone is trying to call `PreparedStatement.executeQuery(sql)`, they're doing something wrong in the first place...

jonathanswenson: Missed that.... great. Assuming that this BQFeaturNotSupported exception is a SQL Exception then I think that is fine.

Maybe it deserves a different exception type/name since isn’t supported by the spec  (but saying it isn’t supported is probably ok).

src/test/java/BQJDBC/QueryResultTest/PreparedStatementTests.java
```diff
@@ -140,6 +140,59 @@ public void ParameterlessTest() {
...
+                java.sql.Types.DOUBLE,
+                java.sql.Types.BOOLEAN,
+                java.sql.Types.BIGINT,
+                java.sql.Types.VARCHAR,
+                java.sql.Types.TIMESTAMP,
+                java.sql.Types.DATE,
+        };
+
+        for (int i = 0; i < queries.length; i ++) {
+            System.out.println("Running query: " + queries[i]);
```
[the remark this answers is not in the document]
kaibankslooker: If you'd like me to! There are a bunch of other tests in that file that `System.out.print` the queries they'll run.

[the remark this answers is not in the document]
kaibankslooker: I figured I'd leave in the `println` just for easier debugging while running these tests, and the others follow the same pattern - would you prefer to avoid `print`s in the test files?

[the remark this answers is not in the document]
jonathanswenson: Maybe just a personal preference. Typically, I try to avoid them.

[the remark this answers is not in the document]
kaibankslooker: Fair enough. I'll go through and remove this and other print statements in this test file.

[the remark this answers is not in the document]
jonathanswenson: 👍

PR #36, opened 2019-11-12
Set log4j root logger level to ERROR to quiet down logs

kaibankslooker:
The default log4j configuration is evaluating to `INFO`. Explicitly set the new log level to `ERROR` by adding a new `src/main/resources/log4j.properties`.


kaibankslooker:
Edit: closing for now until I find a better way to alter the log4j root level at the consumer level.

PR #38, opened 2020-01-28
Add metadata support for DATETIME data type

kaibankslooker:
Note that `DATETIME` is not a native `java.sql.Types` type, so just categorize it as a `TIMESTAMP` when retrieving column types for `DATETIME`s.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQResultSetMetadata.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQResultSetMetadata.java
@@ -305,6 +305,9 @@ class BQResultsetMetaData implements ResultSetMetaData {
         if (Columntype.equals("DATE")) {
             return 10;
         }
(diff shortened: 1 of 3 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQResultSetMetadata.java
```diff
@@ -240,6 +240,10 @@ public int getColumnType(int column) throws SQLException {
             return java.sql.Types.DATE;
         }
 
+        if (Columntype.equals("DATETIME")) {
+            return java.sql.Types.TIMESTAMP;
```
jonathanswenson: Yeah I think this is what other databases that support datetime do (mysql).

kaibankslooker:
Hm - I'll add `precision`. `ColumnTypeName` and `scale` are already handled.

PR #39, opened 2020-02-25
Switch driver to run with synchronous jobs.query endpoint

stalbot:
Had to do a bit of refactoring because parts of the code assume they have a BigQuery API `GetQueryResultsResponse` object. Instead, we need to have those parts of the code take a list of rows and a schema, so that when we get all the rows we need back from the synchronous API, which will be very frequently true since the limit there is 10MB, we don't want to have to go fetch more rows from BQ.

### DIFF
--- a/src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
+++ b/src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
@@ -593,10 +593,6 @@ public class Oauth2Bigquery {
 
         public void setOauthToken(String oauthToken) { this.oauthToken = oauthToken; }
 
-        public String getUserAgent() {
-            return this.userAgent;
-        }
-
         public String getOauthToken() {return this.oauthToken; }
 
         @Override
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -69,7 +69,7 @@ public class BQConnection implements Connection {
 
     private Long maxBillingBytes;
 
-    private boolean useQueryApi = false;
+    private boolean useQueryApi;
 
     private final Set<BQStatementRoot> runningStatements = Collections.synchronizedSet(new HashSet<BQStatementRoot>());
 
@@ -164,8 +164,8 @@ public class BQConnection implements Connection {
         String withServiceAccountParam = caseInsensitiveProps.getProperty("withserviceaccount");
         Boolean serviceAccount = (withServiceAccountParam != null) && Boolean.parseBoolean(withServiceAccountParam);
 
-        String useQueryApiParam = caseInsensitiveProps.getProperty("useQueryApi");
-        useQueryApi = (useQueryApiParam != null) && Boolean.parseBoolean(useQueryApiParam);
+        String useQueryApiParam = caseInsensitiveProps.getProperty("usequeryapi");
+        useQueryApi = Boolean.parseBoolean(useQueryApiParam);
 
         // extract transformQuery property
         String transformQueryParam = caseInsensitiveProps.getProperty("transformquery");
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -1053,7 +1053,7 @@ public class BQConnection implements Connection {
      *
      * This API is faster by avoiding the async overhead of job polling, but does not currently support maxBillingGBs.
      * */
-    boolean getUseQueryApi() {
+    boolean shouldUseQueryApi() {
         return useQueryApi;
     }
 }
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -166,6 +166,7 @@ public class BQConnection implements Connection {
         String withServiceAccountParam = caseInsensitiveProps.getProperty("withserviceaccount");
         Boolean serviceAccount = (withServiceAccountParam != null) && Boolean.parseBoolean(withServiceAccountParam);
 
+        // There's really no reason anyone would ever not want to use this... but leave it in for a bit as a bailout switch
         String useQueryApiParam = caseInsensitiveProps.getProperty("usequeryapi");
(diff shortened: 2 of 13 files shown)


src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -593,10 +593,6 @@ public void setUserAgent(String userAgent) {
 
         public void setOauthToken(String oauthToken) { this.oauthToken = oauthToken; }
 
-        public String getUserAgent() {
```
stalbot, line 596: It seems like this overrides a method that is not final in an upgraded version of the google library. In any case, something yelled at me when I tried to use this and this method is totally unused, so I just removed it

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -1053,4 +1059,13 @@ public int cancelRunningQueries() {
     public Long getMaxBillingBytes() {
         return maxBillingBytes;
     }
+
+    /**
+     * Returns true if queries on this connection should use the synchronous jobs.query api to run queries.
+     *
+     * This API is faster by avoiding the async overhead of job polling, but does not currently support maxBillingGBs.
```
stalbot, line 1066: whoops comment out of date

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -122,33 +124,71 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
...
-            referencedJob = startQuery(querySql, unlimitedBillingBytes);
+            if (this.connection.shouldUseQueryApi()) {
+                QueryResponse qr = BQSupportFuncts.runSyncQuery(
+                        this.connection.getBigquery(),
+                        this.ProjectId,
+                        querySql,
+                        connection.getDataSet(),
+                        this.connection.getUseLegacySql(),
+                        !unlimitedBillingBytes ? this.connection.getMaxBillingBytes() : null,
+                        (long) 10 * 1000, // we need this to respond fast enough to avoid any socket timeouts
```
stalbot, line 140: I didn't dive deeply into what was happening here. Downstream tests that used this library started failing on really long queries without this fix when initially I had the value set to `this.queryTimeout * 1000`, and it seems reasonable to only ask this first request to wait for so long, then we can start polling.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -665,6 +665,30 @@ public static Properties readFromPropFile(String filePath)
         return properties;
     }
 
+    static QueryResponse runSyncQuery(Bigquery bigquery, String projectId,
```
stalbot, line 668: I guess this merits a doc comment

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
```diff
@@ -60,7 +62,7 @@
     BQConnection connection;
 
     /** Variable that stores the set query timeout */
-    int querytimeout = Integer.MAX_VALUE;
+    int querytimeout = Integer.MAX_VALUE / 1000 - 1;
```
stalbot, line 65: This ends up getting multiplied by 1000 somewhere, so the old value was "too big"

src/main/java/net/starschema/clouddb/jdbc/BQScrollableResultSet.java
```diff
@@ -200,6 +180,16 @@ public Object getObject(int columnIndex) throws SQLException {
                     long val = new BigDecimal(result).longValue() * 1000;
                     return new Timestamp(val);
                 }
+                if (Columntype.equals("DATETIME")) {
```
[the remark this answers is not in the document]
kaibankslooker: Now that you mention it, is there a way that we could pull this `Columntype` parsing out into a method that's shared across resultset types?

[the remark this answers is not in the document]
kaibankslooker: Also, shouldn't `getObject`'s types be 1:1 with the types available in the resultset metadata? I see room for problems whenever the metadata identifies a type that then cannot be parsed/converted by `getObject`. 

https://github.com/jonathanswenson/starschema-bigquery-jdbc/blob/573890b161f863bf31295441b78f17a7dab75206/src/main/java/net/starschema/clouddb/jdbc/BQResultSetMetadata.java#L211

Maybe not an action item for this particular PR, but something to think about.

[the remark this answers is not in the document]
stalbot: ^^ I think that's probably right. But I might indeed take it as an action item for later.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -164,6 +166,10 @@ public BQConnection(String url, Properties loginProp) throws SQLException {
         String withServiceAccountParam = caseInsensitiveProps.getProperty("withserviceaccount");
         Boolean serviceAccount = (withServiceAccountParam != null) && Boolean.parseBoolean(withServiceAccountParam);
 
+        // There's really no reason anyone would ever not want to use this... but leave it in for a bit as a bailout switch
+        String useQueryApiParam = caseInsensitiveProps.getProperty("usequeryapi");
+        useQueryApi = useQueryApiParam == null || !useQueryApiParam.equalsIgnoreCase("false");
```
lameyer, line 171: Does `!useQueryApiParam.equalsIgnoreCase("false")` not cover the null case as well?

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
lameyer: lol

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
stalbot: Oh good point it better.

stalbot: Oh, whoops, except it's a null pointer exception, that's why this was there.

PR #40, opened 2020-03-09
Json auth in props support

stalbot:
To get this new test to work, you will need a JSON auth file and a protectedaccount-json.properties file locally.

### DIFF
--- a/src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
+++ b/src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
@@ -58,6 +58,7 @@ import java.security.PrivateKey;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 public class Oauth2Bigquery {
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/starschema/clouddb/cmdlineverification/Oauth2Bigquery.java
```diff
@@ -265,11 +260,13 @@ public static Bigquery authorizeviaservice(String serviceaccountemail,
...
-                                               Integer connectTimeout,
-                                               Integer readTimeout) throws GeneralSecurityException, IOException {
+                                               String jsonAuthContents, Integer readTimeout, Integer connectTimeout) throws GeneralSecurityException, IOException {
         GoogleCredential credential;
         // Determine which keyfile we are trying to authenticate with.
-        if (Pattern.matches(".*\\.json$", keypath)) {
+        if (jsonAuthContents != null) {
+            credential = Oauth2Bigquery.createJsonCredential(jsonAuthContents);
+        } else if (Pattern.matches(".*\\.json$", keypath)) {
+            // For backwards compat: this is no longer the preferred path for JSON (better to use [jsonAuthContents]
```
kaibankslooker: Makes sense to keep the backward-compatible `keypath` codepath around, but won't the next line fail? `createJsonCredential`, if passed a keypath instead of a string representing auth contents, will no longer `File jsonKey = new File(keypath);` before trying to build its inputstream, right? 

Maybe I'm missing some Java magic that will look for a file in that keypath before trying to build an inputstream from it

stalbot: Oh 👍 of course nice catch. I don't think there's actually a test for the json file as keypath code...

PR #42, opened 2020-03-27
Fix getString for TIMESTAMP types

stalbot:
Also moved language compatibility to Java 8 and so bumped to 2.0...

maybe bump to 2.0 is too aggressive? But I was _not_ going to use the Java legacy Date API to do this. And it's time for this project to be on 8.

As to the bug, prior to this change, since TIMESTAMPs come back in JSON over the wire as stringified BigInts, and getString was doing nothing special to handle that, the result for `SELECT CURRENT_TIMESTAMP` was something like `"1.234932532E9"`

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
@@ -42,9 +42,12 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.sql.*;
 import java.sql.Date;
+import java.sql.*;
 import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.*;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -90,6 +93,10 @@
      * The -1 is needed because of the while(Result.next() == true) { } iterating method*/
     private int Cursor = -1;
 
+    private final DateTimeFormatter TIMESTAMP_FORMATTER = DateTimeFormatter
+            .ofPattern("yyyy-MM-dd HH:mm:ss")
```
jonathanswenson, line 97: Do we not want this to have any extra precision here? I know other jdbc drivers have fractional seconds in their base implementation. I don't know if there is any standard way of doing this though.

stalbot: Ah, that's a good point. I was inspired to make this change off of work where all I did for a day or two was stare at what various dialects sent back from `.toString()` on a timestamp, and yet somehow I don't know the answer. Will check it out and report back.

PR #46, opened 2020-04-15
Type parity between BQ and Starschema

kaibankslooker:
Add `TIME` and `BYTES` as supported types. Also add full list of BQ types for reference.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQResultSetMetadata.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQResultSetMetadata.java
@@ -27,12 +27,7 @@
 
 package net.starschema.clouddb.jdbc;
 
-import java.sql.Date;
-import java.sql.ResultSet;
-import java.sql.ResultSetMetaData;
-import java.sql.SQLException;
-import java.sql.Struct;
-import java.sql.Timestamp;
+import java.sql.*;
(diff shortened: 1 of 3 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQResultSetMetadata.java
```diff
@@ -192,16 +187,24 @@ public String getColumnName(int column) throws SQLException {
...
+     * NUMERIC => java.sql.Types.NUMERIC<br>
+     * FLOAT => java.sql.Types.DOUBLE<br>
+     * BOOLEAN => java.sql.Types.BOOLEAN<br>
+     * STRING => java.sql.Types.VARCHAR<br>
+     * BYTES => java.sql.Types.VARCHAR<br>
+     * DATE => java.sql.Types.DATE<br>
+     * DATETIME => java.sql.Types.TIMESTAMP<br>
+     * TIME => java.sql.Types.TIME<br>
+     * TIMESTAMP => java.sql.Types.TIMESTAMP<br>
+     * ARRAY => unsupported<br>
```
kaibankslooker, line 201: Note that arrays seem to be natively supported (selecting an ARRAY in the sql runner works fine)... and GEOGRAPHY just doesn't seem worth implementing.

PR #50, opened 2020-07-28
Implement basic driver support for ARRAY types

kaibankslooker:
Introduce new `getColumnMode(int index)` accessor on `BQResultSetMetadata` to allow consumers to determine whether results are arrays or not. 

Also change behavior to return raw string values for "REPEATED"-mode values (arrays) so that consumers can interpret/convert types as needed. 

Note that `getObject()` and `getString()` will behave identically for arrays: both will return the raw string value of the result.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,5 +154,5 @@
     </plugins>
     <finalName>${project.artifactId}-${project.version}-thin</finalName>
   </build>
-  <version>2.1.2</version>
+  <version>2.1.3</version>
 </project>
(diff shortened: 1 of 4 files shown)


kaibankslooker:
@stalbot here's a new PR containing the changes we discussed on Slack. Arrays will now be returned as parseable JSON strings by both `getString()` and `getObject()`, and the new `getColumnMode(int index)` accessor allows us (in HT) to determine when we're receiving array results from the driver.

PR #53, opened 2020-08-04
Fix sync query api path for non-US/Eu locations

stalbot:
This path was broken for any query that did not fetch all the rows in the initial load in non-US/EU locations, because it did not call `setLocation` when getting the job.

### DIFF
--- a/src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
+++ b/src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
@@ -37,6 +37,7 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Arrays;
 //import net.starschema.clouddb.bqjdbc.logging.Logger;
 
 /**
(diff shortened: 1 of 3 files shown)


src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
```diff
@@ -581,4 +582,36 @@ public void QueryResultTestTokyo01() {
...
+        final String sql = "SELECT meaning FROM tokyo_star.meaning_of_life;";
+        String[][] expectation = new String[][]{ {"42"} };
+
+        this.logger.info("Test Tokyo number: 1");
+        this.logger.info("Running query:" + sql);
+
+        java.sql.ResultSet Result = null;
+        try {
+            Statement s = QueryResultTest.con.createStatement();
+            ((BQStatement) s).setQueryApiInitialFetchMaxRows((long) 0);
```
stalbot, line 597: this oddity comes from the fact that the only table we easily have access to for these tests has... exactly one row. And, for obvious reasons, you are not allowed to set maxRows on Statement to zero. So we make this custom API and use it for the test. In theory someone else in the future might want to use it for something.

stalbot: ... nevermind. Looked at this again and remembered how to SQL

PR #54, opened 2020-09-08
Fixup the way the sync query code handles cancelling a job by reference

stalbot:
Cancel code relied on existence of a `job` on `BQStatement`, but that was only set (only could be set) by the async code. We need to wait for the sync code to time out, then cancel the query. 

Without further change here, this means that the absolute fastest a query can be cancelled is ten seconds. We can reduce the ten seconds (it's just a constant we control) but that means more queries would miss on the sync endpoint and have to poll, so it's a bit of a trade-off. For now, will reduce the sync time to five seconds in this PR. We could also enable that to be controlled by a connection parameter in the future for folks who really want to tune, but don't bother for right now because it's not clear at five seconds plus how much benefit the sync endpoint gives you anyway.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,5 +154,5 @@
     </plugins>
     <finalName>${project.artifactId}-${project.version}-thin</finalName>
   </build>
-  <version>2.1.5</version>
+  <version>2.1.6</version>
 </project>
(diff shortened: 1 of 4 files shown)


kaibankslooker (approved):
Looks good to me. Implementation/setup/cleanup of the `runningSyncThread` and `lastSyncResponse` variables seems solid.

PR #55, opened 2020-10-07
Fix cancellation when the cancel comes after the sync part of the query has completed

stalbot:
The old code was using the existence of the `currentlyRunningSyncThread` to check if we were in the sync case, but that thread was nulled out after the sync path completed! So make sure we also check for a `lastSyncResponse`.


kaibankslooker (approved):
Makes sense, looks good!

PR #57, opened 2020-10-08
Add support for DML (insert / delete / update)

fzakaria:
BigQuery Standard SQL now supports DELETE / INSERT / UPDATE where it previously did not through the legacy SQL.

I noticed when using **sequel** with our BigQuery JDBC driver that sequel will correctly try to call [executeUpdate](https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#executeUpdate(java.lang.String)) when trying to do DML.

The current implementation unfortunately simply returns an _UnsupportedException_.
Add support for executing the query and returning the number of rows affected.


src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
```diff
@@ -240,6 +241,73 @@ public boolean execute(String arg0, String[] arg1) throws SQLException {
...
 
+    /**
+     * Execute DML (DELETE, INSERT, UPDATE). If you want to perform a SELECT or other DDL
+     * look at {@link #execute(String)}
+     */
+    private int executeDML(String sql) throws SQLException {
+        if (this.isClosed()) {
+            throw new BQSQLException("This Statement is Closed");
+        }
+        if (!connection.shouldUseQueryApi()) {
```
fzakaria, line 252: For simplicity i only added this feature for the query API.
The documentation in the driver seems like it's not sensible to use the alternate form anyways any longer.

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
```diff
@@ -240,6 +241,73 @@ public boolean execute(String arg0, String[] arg1) throws SQLException {
...
+                                        referencedJob).getNumDmlAffectedRows()
+                            );
+                }
+                Thread.sleep(500);
+                this.logger.debug("slept for 500" + "ms, querytimeout is: "
+                        + this.querytimeout + "s");
+            } while (System.currentTimeMillis() - this.starttime <= (long) this.querytimeout * 1000);
+        } catch (IOException | InterruptedException e) {
+            throw new BQSQLException("Something went wrong with the query: " + sql, e);
+        }
```
fzakaria, line 304: TODO: Probably can wrap this functionality as it's used a ton in the codebase into a helper method.

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
```diff
@@ -197,7 +198,7 @@ public boolean execute(String arg0, boolean unlimitedBillingBytes) throws SQLExc
      * @throws BQSQLFeatureNotSupportedException
      */
 
-    public boolean execute(String arg0, int arg1) throws SQLException {
+    public boolean execute(String sql, int autoGeneratedKeys) throws SQLException {
```
fzakaria, line 201: Just added the names of the arguments so it's more legible according to the javadocs @ https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html

pom.xml
```diff
@@ -154,5 +155,5 @@
     </plugins>
     <finalName>${project.artifactId}-${project.version}-thin</finalName>
   </build>
-  <version>2.1.7</version>
+  <version>2.1.8-SNAPSHOT</version>
```
stalbot, line 158: Why this "-SNAPSHOT"? I mean, there is _something_ of a "release" process for this, it's not all local builds.

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
stalbot: Is this a TODO for now or later?

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
stalbot: Well, when we initially rolled it out it had a couple bugs, and some folks put the fallback in their JDBC params, so some people might still have `!shouldUseQueryApi()`. But that being said, it was really only a fallback for the existing code, you could probably just ignore the flag here and do the one true code path.

pom.xml
stalbot: Also nit I'd say make this 2.2.0 since it's a net new feature

stalbot (commented):
Generally seems good! I don't think cancelling will work on this method since it doesn't do the same dance as `execute` to remember the current job, but I wonder if we want to support cancelling DDL anyway. A quick internet search doesn't yield any obvious answers.

pom.xml
stalbot: oic this is an explicitly WIP gesture on your part, carry on carry on

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
fzakaria: Removed!

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
fzakaria: This is a TODO for later. This pattern is used a little bit and should be cleaned up:
```bash
❯ rg "while \(System.currentTimeMillis\(\)"
src/main/java/net/starschema/clouddb/jdbc/BQPreparedStatement.java
283:            while (System.currentTimeMillis() - this.starttime <= (long) this.querytimeout * 1000);

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
298:            } while (System.currentTimeMillis() - this.starttime <= (long) this.querytimeout * 1000);
385:            while (System.currentTimeMillis() - this.starttime <= (long) this.querytimeout * 1000);

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
223:            while (System.currentTimeMillis() - this.starttime <= (long) this.querytimeout * 1000);
```

This JDBC driver in fact can use _a lot of love_ especially given that the state of the BigQuery API has changed since 2015. It might make sense to make a branch that removes legacy features as well and have our standard SQL use that instead.

(A more interesting idea would be to either migrate to the Samba JDBC driver or I think we could write one ourselves better suited for our needs)

pom.xml
fzakaria: Bumped to 2.2.0 but keeping SNAPSHOT I guess?
Once this lands, I can open a PR or you can let me know about the release process to bump it to non-SNAPSHOT.

fzakaria:
@stalbot not sure if you noticed... but I didn't add a unit test for this.
What's your feedback on that here? It's not as simple as SELECT unit test that reads from publicdata; testing DML would entail having credentials that can create a dataset -> table -> insert.

I can definitely add one but not sure given that the tests don't even run anyways.
(I did test this however via Looker & PDT codebase)

stalbot (approved):
As you say, a functional test for this would require paying credentials on BigQuery. Better to just, with this isolated function, manually test it, or test it via some other client, if you change it.

fzakaria:
BigQuery Standard SQL now supports DELETE / INSERT / UPDATE where it previously did not through the legacy SQL.

I noticed when using sequel with our BigQuery JDBC driver that sequel will correctly try to call executeUpdate when trying to do DML.

The current implementation unfortunately simply returns an UnsupportedException.
Add support for executing the query and returning the number of rows affected.

PR #58, opened 2020-10-08
Added github action

fzakaria:
:+1: Big thanks to @jonathanswenson for adding the secrets necessary to run this.

Testing, testing, testing
We now have testing for our jdbc driver; this repository is a bit **less** humiliating.

Turns out some of the tests are in fact broken because they rely on a certain ordering :+1: 
At least now we can run the tests more often and discover these broken ones.

image

### DIFF
--- a/src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
+++ b/src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
@@ -30,6 +30,7 @@ import net.starschema.clouddb.jdbc.BQStatement;
 import net.starschema.clouddb.jdbc.BQSupportFuncts;
 import net.starschema.clouddb.jdbc.BQSupportMethods;
 import org.apache.log4j.Logger;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


fzakaria:
@jonathanswenson @stalbot we can have a GitHub action to also deploy to GitHub packages instead of Nexus fairly easily
https://docs.github.com/en/free-pro-team@latest/actions/guides/publishing-java-packages-with-maven#publishing-packages-to-github-packages

jonathanswenson (approved):
👍 Thank you for doing this!

src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
```diff
@@ -110,6 +111,12 @@ public void NewConnection() {
         NewConnection(null);
     }
 
+    @After
+    public void TeardownConnection() throws SQLException {
+        QueryResultTest.con.close();
+        QueryResultTest.con = null;
+    }
+
```
fzakaria, line 119: Depending on the order, the connection was being reused and it may have been created in legacy or non-legacy mode.
Certain tests used the legacy way of referring to projects
```sql
SELECT TOP(word, 10), COUNT(*) FROM publicdata:samples.shakespeare
```

jonathanswenson (approved):

PR #59, opened 2020-10-09
Add release workflow

fzakaria:
This release workflow runs through Maven's release plugin which does the
following:

## Prepare
1. Change the version in the POM from x-SNAPSHOT to a new version
2. Transform the SCM information in the POM to include the final destination of the tag
3, Commit the modified POM
4. Tag the code in the SCM with a version name
5. Bump the version in the POM to a new value y-SNAPSHOT
6. Commit the modified POM

## Perform
1. Checkout from an SCM URL with optional tag
2. Run the predefined Maven deploy goal

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: Publish package to GitHub Packages
+name: Publish Package
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


fzakaria:
Not 100% if this will work; might take some trial and error.

The goal would be to add a final step to also publish to OSSRH (nexus); which would be just another deploy step.

fzakaria:
Not 100% if this will work; might take some trial and error.

The goal would be to add a final step to also publish to OSSRH (nexus); which would be just another deploy step.

PR #63, opened 2020-10-15
Make the sync response null when the query completes

stalbot:
There is some undesirable behavior where in the sync query mode, we will cancel the job on `Statement#close` even though the job is already done. BigQuery happily tells us "this job is already done" and moves on, but no point initiating an API request to BQ in the general `stmt.close` case (surely slows down queries), so let's make sure to clean up after ourselves by nulling out `syncResponseFromCurrentQuery` (and change its name so that it's more obviously something that only stays around for the current query's lifecycle).

This is ultimately a bug, in that if you don't have permission to execute a cancel, this means that the normal flow where you make a Statement, execute a query, and then close the statement throws an exception, so add a test hat we are not calling cancel at the end of a successful query run.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
@@ -49,7 +49,7 @@ public class BQStatement extends BQStatementRoot implements java.sql.Statement {
     public static final int MAX_IO_FAILURE_RETRIES = 3;
     private Job job;
     private AtomicReference<Thread> runningSyncThread = new AtomicReference<>();
-    private AtomicReference<QueryResponse> lastSyncResponse = new AtomicReference<>();
+    private AtomicReference<QueryResponse> syncResponseFromCurrentQuery = new AtomicReference<>();
 
     /**
      * Enough time to give fast queries time to complete, but fast enough that if we want to cancel the query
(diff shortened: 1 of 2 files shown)


fzakaria:
@stalbot what about checking in cancel for `maybeAlreadyComplete.jobComplete()` as well?

```  /**
   * Whether the query has completed or not. If rows or totalRows are present, this will always be
   * true. If this is false, totalRows will not be available.
   * The value may be {@code null}.
   */
  @com.google.api.client.util.Key
  private java.lang.Boolean jobComplete;
```

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -105,6 +105,7 @@ public ResultSet executeQuery(String querySql) throws SQLException {
         }
         finally {
             this.job = null;
+            this.syncResponseFromCurrentQuery.set(null);
```
fzakaria, line 108: This seems to be the relevant line.

fzakaria (approved):
LGTM.
At some point we should add more tests & have them not rely on Looker specific datasets.
(Looking at myself in the mirror)

ex.
```stmt.executeQuery("SELECT * FROM orders limit 1");```

PR #67, opened 2020-10-23
Refactor, cleanup and simplify the code

fzakaria:
This may look like a _large change_, but it's largely deleting dead code and simplifying the dependencies.
I will call out within the review anything that's particularly meaningful but it should not affect much.

The largest change is removing the oauth flow within the jdbc driver itself which required it to have a dependency on Jetty.
It's not clear on the use case for this since it relied on printing to System.out the web URL to visit to start.

- deleted old Eclipse files that should never have been committed.
- fixed references to new repository name --> bqjdbc
- removed GUI code that requires javaFX
- removed package name cmdlineverification
- consolidated the code within the jdbc package
- deleted all the dead oauth code flow
- changed logging to SLF4J


README.md
```diff
@@ -1,6 +1,7 @@
 # bqjdbc
 
 ![Maven Central](https://img.shields.io/maven-central/v/com.github.jonathanswenson/bqjdbc)
+[![javadoc](https://javadoc.io/badge2/com.github.jonathanswenson/bqjdbc/javadoc.svg)](https://javadoc.io/doc/com.github.jonathanswenson/bqjdbc)
```
fzakaria, line 4: This is a really sweet service~!

jonathanswenson (approved):
👍 Glad to see these old systems going away. antlr next!

PR #70, opened 2020-10-26
Implement Schema Methods

aryeh-looker:
Some `dbcp2` expects `(get|set)Scehma` to work if a schema is set on the connection, here we implement.

From `dbcp2`:
image


fzakaria (approved):
LGTM

PR #71, opened 2020-10-26
Remove support for Catalog -> ProjectId name mangling

kalenp:
A prior version of this driver supported translating between SQL-92 and
BigQuery Legacy dialects.  As part of this support, it would mangle
BigQuery Project Ids like `looker.com:sample-project` to
`looker_com__sample-project`.  We found that there were some bugs in
this translation leading to the mangled name being sent to BigQuery when
the real ProjectId should have been sent.

We recently removed support for this translation between dialects.
Since we no longer require that, we can now remove most of the code
which did the name mangling.

In this PR, we fix on using the Project Id internally as much as
possible.  Only in public methods of the JDBC API do we accept a Catalog
name.  All internal methods translate this to a Project Id before doing
any further processing of them.

We do retain one vestige of the original name mangling.  We continue to
support the mangled names in the JDBC URI.  This is to maintain
compatibility with an existing configurations which used the mangled
name.  Should we decide to drop this support in the future, it should be
trivial now that all of the mangling is contained in the single
`CatalogName.toProjectId` method.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -59,9 +59,10 @@ public class BQConnection implements Connection {
     private String dataset = null;
 
     /**
-     * The projectid which needed for the queries.
+     * The ProjectId for the connection
      */
     private String projectId = null;
+
     /** Boolean to determine if the Connection is closed */
     private boolean isclosed = false;
 
@@ -114,16 +115,14 @@ public class BQConnection implements Connection {
             Matcher matchData = projectAndDatasetMatcher.matcher(pathParams);
 
             if (matchData.find()) {
-                this.projectId = matchData.group(1);
+                this.projectId = CatalogName.toProjectId(matchData.group(1));
                 this.dataset = matchData.group(2);
             } else {
-                this.projectId = pathParams;
+                this.projectId = CatalogName.toProjectId(pathParams);
             }
         } catch (UnsupportedEncodingException e1) {
             throw new BQSQLException(e1);
         }
-        // lets replace the : with __ and . with _
-        this.projectId = this.projectId.replace(":", "__").replace(".", "_");
 
         Properties caseInsensitiveLoginProps = new Properties();
 
@@ -227,7 +226,7 @@ public class BQConnection implements Connection {
         else {
             throw new IllegalArgumentException("Must provide a valid mechanism to authenticate.");
         }
-        logger.debug("The project id for this connections is: " + this.projectId);
+        logger.debug("The project id for this connections is: " + projectId);
     }
 
     /**
@@ -371,7 +370,7 @@ public class BQConnection implements Connection {
         }
         logger.debug("Creating statement with resultsettype: forward only," +
                 " concurrency: read only");
-        return new BQStatement(this.projectId, this);
+        return new BQStatement(projectId, this);
     }
 
     /** {@inheritDoc} */
@@ -383,8 +382,7 @@ public class BQConnection implements Connection {
         }
         logger.debug("Creating statement with resultsettype: " + resultSetType
                 + " concurrency: " + resultSetConcurrency);
-        return new BQStatement(this.projectId, this, resultSetType,
-                resultSetConcurrency);
+        return new BQStatement(projectId, this, resultSetType, resultSetConcurrency);
     }
 
     /** {@inheritDoc} */
@@ -417,8 +415,8 @@ public class BQConnection implements Connection {
     }
 
     @Override
-    public String getSchema() {
-        return this.dataset;
+    public String getSchema() throws SQLException {
+      return getDataSet();
     }
 
     public void abort(Executor executor) throws SQLException {
@@ -464,7 +462,7 @@ public class BQConnection implements Connection {
     @Override
     public String getCatalog() throws SQLException {
         logger.debug("function call getCatalog returning projectId: " + projectId);
-        return this.projectId;
+        return projectId;
     }
 
     /**
@@ -526,7 +524,7 @@ public class BQConnection implements Connection {
      * Getter method for projectId
      */
     public String getProjectId() {
-        return this.projectId;
+        return projectId;
     }
 
     /**
@@ -638,7 +636,7 @@ public class BQConnection implements Connection {
                             + String.valueOf(timeout));
         }
         try {
-            this.bigquery.datasets().list(this.projectId.replace("__", ":").replace("_", ".")).execute();
+            this.bigquery.datasets().list(projectId).execute();
         } catch (IOException e) {
             return false;
         }
@@ -729,10 +727,9 @@ public class BQConnection implements Connection {
     @Override
     public PreparedStatement prepareStatement(String sql) throws SQLException {
         this.logger.debug("Creating Prepared Statement project id is: "
-                + this.projectId + " with parameters:");
+                + projectId + " with parameters:");
         this.logger.debug(sql);
-        PreparedStatement stm = new BQPreparedStatement(sql, this.projectId,
-                this);
+        PreparedStatement stm = new BQPreparedStatement(sql, projectId, this);
         return stm;
     }
 
@@ -756,12 +753,12 @@ public class BQConnection implements Connection {
     public PreparedStatement prepareStatement(String sql, int resultSetType,
                                               int resultSetConcurrency) throws SQLException {
         this.logger.debug("Creating Prepared Statement" +
-                " project id is: " + this.projectId +
+                " project id is: " + projectId +
                 ", resultSetType (int) is: " + String.valueOf(resultSetType) +
                 ", resultSetConcurrency (int) is: " + String.valueOf(resultSetConcurrency)
                 + " with parameters:");
         this.logger.debug(sql);
-        PreparedStatement stm = new BQPreparedStatement(sql, this.projectId,
+        PreparedStatement stm = new BQPreparedStatement(sql, projectId,
                 this, resultSetType, resultSetConcurrency);
         return stm;
     }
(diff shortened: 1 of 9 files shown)


src/test/java/BQJDBC/QueryResultTest/JdbcUrlTest.java
```diff
@@ -41,10 +41,22 @@ public void urlWithDefaultDatasetShouldWork() throws SQLException {
     @Test
     public void projectWithColons() throws SQLException {
         String urlWithColonContainingProject = URL.replace(bq.getProjectId(), "before:after");
+        try {
+            BQConnection bq_with_colons = new BQConnection(urlWithColonContainingProject, new Properties());
```
jonathanswenson, line 45: nit: snake_case?

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -277,20 +211,22 @@ public static String getDatasetIdFromTableDotGetId(String getId) {
...
                 ", projectId: " + (projectId != null ? projectId : "null"));
-        projectId = projectId.replace("__", ":").replace("_", ".");
         List<Datasets> datasetcontainer = connection.getBigquery().datasets()
                 .list(projectId).execute().getDatasets();
 
         if (datasetcontainer != null && datasetcontainer.size() != 0) {
             if (datasetname != null) {
                 List<Datasets> datasetsSearch = new ArrayList<Datasets>();
                 for (Datasets in : datasetcontainer) {
-                    if (matchPattern(getDatasetIdFromDatasetDotGetId(in.getId()),
```
jonathanswenson, line 293: I'm just curious how this function name happened.

src/main/java/net/starschema/clouddb/jdbc/CatalogName.java
```diff
@@ -0,0 +1,47 @@
...
+ * accepted an encoded version of the ProjectId as the Catalog name.  So as not to break existing
+ * clients, we will still accept this encoding as part of the JDBC URL identifier.
+ *
+ * The encoding affects Projects which are prefixed with a domain, such as
+ * "looker.com:sample-project".  These may be selected with a Catalog name of
+ * "looker_com__sample-project", where the period (`.`) is replaced wih a single underscore (`_`)
+ * and the colon (`:`) is replaced with two underscores (`__`).
+ */
+public final class CatalogName {
+  private final String catalog;
```
jonathanswenson, line 16: Not sure why, but this has 2 spaces, rather than 4.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -464,7 +462,7 @@ public Bigquery getBigquery() {
     @Override
     public String getCatalog() throws SQLException {
         logger.debug("function call getCatalog returning projectId: " + projectId);
-        return this.projectId;
```
jonathanswenson, line 467: 👍

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -114,16 +115,14 @@ public BQConnection(String url, Properties loginProp) throws SQLException {
             Matcher matchData = projectAndDatasetMatcher.matcher(pathParams);
 
             if (matchData.find()) {
-                this.projectId = matchData.group(1);
+                this.projectId = CatalogName.toProjectId(matchData.group(1));
```
jonathanswenson, line 118: 👍

jonathanswenson (approved):

fzakaria:
nit: it wasn't `A prior version of this driver supported translating between Legacy and
Standard BigQuery dialects.` but rather SQL-92 -> BigQuery Legacy

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -227,7 +226,7 @@ else if (oAuthAccessToken != null) {
         else {
             throw new IllegalArgumentException("Must provide a valid mechanism to authenticate.");
         }
-        logger.debug("The project id for this connections is: " + this.projectId);
+        logger.debug("The project id for this connections is: " + projectId);
```
fzakaria, line 229:
```suggestion
        logger.debug("The project id for this connections is {} ", projectId);
```

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -464,7 +462,7 @@ public Bigquery getBigquery() {
     @Override
     public String getCatalog() throws SQLException {
         logger.debug("function call getCatalog returning projectId: " + projectId);
```
fzakaria, line 464:
```suggestion
        logger.debug("function call getCatalog returning projectId: {}", projectId);
```

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -729,10 +727,9 @@ public CallableStatement prepareCall(String sql, int resultSetType,
     @Override
     public PreparedStatement prepareStatement(String sql) throws SQLException {
         this.logger.debug("Creating Prepared Statement project id is: "
-                + this.projectId + " with parameters:");
+                + projectId + " with parameters:");
```
fzakaria, line 730: nit same if you want.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -756,12 +753,12 @@ public PreparedStatement prepareStatement(String sql, int autoGeneratedKeys)
     public PreparedStatement prepareStatement(String sql, int resultSetType,
                                               int resultSetConcurrency) throws SQLException {
         this.logger.debug("Creating Prepared Statement" +
-                " project id is: " + this.projectId +
+                " project id is: " + projectId +
```
fzakaria, line 756: nit same

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -73,23 +73,23 @@ public BQStatement(String projectid, BQConnection bqConnection) {
...
      * @param bqConnection
      * @param resultSetType
      * @param resultSetConcurrency
      * @throws BQSQLException
      */
-    public BQStatement(String projectid, BQConnection bqConnection,
+    public BQStatement(String projectId, BQConnection bqConnection,
                        int resultSetType, int resultSetConcurrency) throws BQSQLException {
-        logger.debug("Constructor of BQStatement is running projectid is: " + projectid +
+        logger.debug("Constructor of BQStatement is running projectId is: " + projectId +
```
fzakaria, line 84: nit

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -158,69 +158,29 @@ public static Properties getUrlQueryComponents(String url, Properties defaults)
         return components;
     }
 
-    /**
```
fzakaria, line 161: :+1:

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -158,69 +158,29 @@ public static Properties getUrlQueryComponents(String url, Properties defaults)
...
-                ProjectReference projRef = projects2.getProjectReference();
-                projRef.setProjectId(projRef.getProjectId().replace(".", "_").replace(":", "__"));
-                projects2.setProjectReference(projRef);
-            }
-            if (catalogName != null) {
+            if (projectId != null) {
                 List<Projects> ProjectsSearch = new ArrayList<Projects>();
                 for (Projects project : projects) {
-                    if (project.getId().contains(catalogName)) {
+                    if (project.getId().equals(projectId)) {
```
fzakaria, line 183: @kalenp I think this method can be rethought.
How can this ever return a List now -- there should only be **1**.

It should either simply return the BigQuery _Project_ model or itself be a no-op now.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -507,47 +411,38 @@ public static String getTableIdFromTableDotGetId(String getId) {
      *                     <p>
      */
     public static List<Table> getTables(BQConnection connection,
-                                        String catalog, String schema, String tablename) throws IOException {
+                                        String projectIdFilter, String schema, String tablename) throws IOException {
```
fzakaria, line 414: Not sure if this is a filter in the same sense anymore as it's doing an **exact** match.
It would be good to explore this as it does not seem 1:1 with the previous intention of the code
(not that the previous intention was right probably)

fzakaria (commented):
I think there are some questions to address.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
kalenp: If `projectId == null`, then this will return the full list of projects.  See https://github.com/jonathanswenson/bqjdbc/pull/71/files/48675ef14ed622ebf9267d694f6d3445c2e4d875#diff-72ff40b7727288cb223d0640acec93efc2e28f70bd903b0b758a99f44a5f5511L233 below.

kalenp:
I would rather address all of the logging issues in a separate cleanup pass, so as not to complicate this PR much more.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
fzakaria: That commit link doesn't work anymore with the force push but the API here is pretty nonsensical

aryeh-looker:
Tag me next time!! 😉

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
aryeh-looker: Neat, so we are _explicitly_ un-mangling here now 👍

PR #73, opened 2020-11-05
Fixup NaN values in double types

stalbot:
Previously, the unnecessary parsing to BigDecimal would explode on NaN (and Inf, etc). Double.parseDouble has no problem with this.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
@@ -215,18 +215,17 @@ public class BQForwardOnlyResultSet implements java.sql.ResultSet {
     /** Parse floating point types */
     private Double toDouble(String value) throws SQLException {
         try {
-            return Double.valueOf(new BigDecimal(value).doubleValue());
+            return Double.parseDouble(value);
         }
         catch (NumberFormatException e) {
             throw new BQSQLException(e);
         }
     }
 
-
     /** Parse integral types */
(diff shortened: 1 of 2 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -215,18 +215,17 @@ private Boolean toBoolean(String value) throws SQLException {
...
             throw new BQSQLException(e);
         }
     }
 
-
     /** Parse integral types */
     private Long toLong(String value) throws SQLException {
         try {
-            return Long.valueOf(new BigDecimal(value).longValue());
+            return Long.parseLong(value);
```
stalbot, line 228: Not technically related, but IntelliJ was giving an "unnecessary boxing" warning and it seemed like a sensible fixup.

kaibankslooker (approved):
LGTM - nice and simple fix. Surprised `Double.parseDouble()` wasn't used in the first place.

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -215,18 +215,17 @@ private Boolean toBoolean(String value) throws SQLException {
     /** Parse floating point types */
     private Double toDouble(String value) throws SQLException {
         try {
-            return Double.valueOf(new BigDecimal(value).doubleValue());
```
fzakaria, line 218: weird

fzakaria (approved):

PR #76, opened 2021-02-09
Add support for default labels to jobs executed

fzakaria:
This is the starting framework to add support to add labels (or tags) to  query requests.
This came out of a discussion with BigQuery engineers as a way to easily identify all payloads originating from Looker services.


fzakaria:
@mkou

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -247,7 +247,8 @@ protected QueryResponse runSyncQuery(String querySql, boolean unlimitedBillingBy
                         this.connection.getUseLegacySql(),
                         !unlimitedBillingBytes ? this.connection.getMaxBillingBytes() : null,
                         SYNC_TIMEOUT_MILLIS, // we need this to respond fast enough to avoid any socket timeouts
-                        (long) getMaxRows()
+                        (long) getMaxRows(),
+                        this.connection.getLabels()
```
mkou, line 251: So from what I understand here you pass on the labels wen creating your DB connection and then use the labels set in the connection and send them with each Query API.
- What if we want to pass on query related information (e.g `queryOrigin: "runner"` or let's imagine a trace ID) to the API, I imagine we are not recreating a connection every time ?

mkou: @fzakaria ^

fzakaria: This will probably need to be extended so that you can pass labels on the actual query -- very similar to maxBillable.
You can probably break it out anyways into two PRs

mkou (approved):
Code looks good

PR #77, opened 2021-02-26
Disable pretty printing by default

fzakaria:
According to feedback from the BigQuery team there is significant overhead in the whitespace processing for BigQuery pretty printed JSON.

They recommend disabling it and they call out bigquery.jobs.getQueryResults and bigquery.tabledata.list as the big winners.

It was kind of tedious to turn it off for every API call, so I introduce a new proxy _Bigquery_ class that sets it for every request type.

### DIFF
--- a/src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
+++ b/src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
@@ -238,7 +238,7 @@ public class QueryResultTest {
             result = stmt.executeQuery(sql);
         } catch (SQLException e) {
             this.logger.debug("SQLexception" + e.toString());
-            Assert.assertTrue(e.toString().contains("Not found: Table measurement-lab:m_lab.2010_01"));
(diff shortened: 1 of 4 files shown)


src/test/java/BQJDBC/QueryResultTest/QueryResultTest.java
```diff
@@ -238,7 +238,7 @@ public void QueryResultTest03() {
             result = stmt.executeQuery(sql);
         } catch (SQLException e) {
             this.logger.debug("SQLexception" + e.toString());
-            Assert.assertTrue(e.toString().contains("Not found: Table measurement-lab:m_lab.2010_01"));
+                Assert.assertTrue(e.toString().contains("Not found: Table guid754187384106:m_lab.2010_01"));
```
fzakaria, line 241: The build was failing here interestingly. How did it pass before?
Clearly the SQL above was "guid754187384106:m_lab.2010_01"

The Exception was just never being thrown ?

stalbot (approved):
LGTM. Maybe ideally we'd add a pedantic unit test about the pretty-printedness of requests, but that's pretty pedantic.

PR #78, opened 2021-02-26
Random improvements for tests

fzakaria:
- Made the java package name for the tests more sensible
- Enabled TimeoutTest (it previously had a lowercase _t_ so it was excluded). Although what's the point of this class ? It's identical to QueryResultTest. :facepalm:


src/test/java/net/starschema/clouddb/jdbc/TimeoutTest.java
```diff
@@ -76,38 +73,26 @@ public void isvalidtest() {
...
+            } catch (Exception e) {
+                this.logger.error("Error in connection" + e.toString());
+                Assert.fail("General Exception:" + e.toString());
             }
-        } catch (SQLException e1) {
-            e1.printStackTrace();
+            this.logger.info(((BQConnection) TimeoutTest.con).getURLPART());
         }
-        try {
-            this.logger.info("thread will sleep for 1 minute");
```
fzakaria, line 104: Why!

jonathanswenson: Your guess is as good as mine 😔

src/test/java/com/google/api/services/bigquery/MinifiedBigqueryTest.java
```diff
@@ -0,0 +1,34 @@
+package com.google.api.services.bigquery;
+
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.testing.http.MockHttpTransport;
+import com.google.api.client.testing.json.MockJsonFactory;
+import org.assertj.core.api.Assertions;
+import org.junit.Test;
+
+public class MinifiedBigqueryTest {
```
fzakaria, line 9: @stalbot followup from previous PR.
Doesn't test return but at least validates we send the request param.

src/test/java/net/starschema/clouddb/jdbc/TimeoutTest.java
```diff
@@ -76,38 +73,26 @@ public void isvalidtest() {
...
-            Thread.sleep(1000 * 1); // 1000milisec = 1 sec * 60 = 1 minute
```
stalbot, line 105: also lol math

stalbot (approved):
obviously this is better than it was before!

src/test/java/net/starschema/clouddb/jdbc/TimeoutTest.java
fzakaria: yea lol, didn't even read that.
I just discounted the whole area mentally lol.

PR #80, opened 2021-03-08
Add support for custom root URL via JDBC parameter

wnob:
Enable using e.g. `rootUrl=https%3A%2F%2Fcustom.bq.com%2F` (URL-decoded: `https://custom.bq.com/`) to hit alternative endpoints to `https://bigquery.googleapis.com/`.

I've added a new properties file for the test and just checked it in since it does not contains any secrets per se (though it references `bigquery_credentials.p12` even though it doesn't make any real requests). I can also just add it to the zip archive that I uploaded to Secret Manager for easier distribution of the files. I also included instructions in the README for accessing those secrets, which are accessible to everyone in the `cloud-looker-model-dev@google.com` group and could potentially be made available to others if we want to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,10 +62,24 @@ A fat (shaded) jar is also available:
 
 We require Java (JDK 8 through 15) and Apache Maven (3.2.5 or higher).
 
-```
+```bash
 $ git clone git://github.com/jonathanswenson/starschema-bigquery-jdbc.git
 $ cd starschema-bigquery-jdbc
 $ mvn clean install -DskipTests
+```
+
+### Local testing
+Testing locally requires access to secret credentials stored in
+[Google Secret Manager](https://console.cloud.google.com/security/secret-manager/secret/bqjdbc-resources-zip/versions?project=346285421363)
+for accessing Looker-owned BigQuery testing tables.
+Once granted access, download and extract the zip file:
+
+```bash
+# Assuming you've already authenticated with `gcloud auth login`:
+$ gcloud config set project secrets-305922
+$ cd src/test/resources
+$ gcloud secrets versions access latest --secret=bqjdbc-resources-zip --format='get(payload.data)' | tr '_-' '/+' | base64 -d > resources.zip
+$ unzip resources.zip
 ```
 
 ### Releases
--- a/README.md
+++ b/README.md
@@ -68,20 +68,6 @@ $ cd starschema-bigquery-jdbc
 $ mvn clean install -DskipTests
 ```
 
-### Local testing
-Testing locally requires access to secret credentials stored in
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -62,12 +62,26 @@ A fat (shaded) jar is also available:
...
+Once granted access, download and extract the zip file:
+
+```bash
+# Assuming you've already authenticated with `gcloud auth login`:
+$ gcloud config set project secrets-305922
+$ cd src/test/resources
+$ gcloud secrets versions access latest --secret=bqjdbc-resources-zip --format='get(payload.data)' | tr '_-' '/+' | base64 -d > resources.zip
+$ unzip resources.zip
+```
+
```
wnob, line 84: Not sure how we feel about adding this to the README. It's not like the fact that running the tests requires secrets is a secret, and it's not like randoms have access to this cloud project, just feels weird linking to it in a public README.

stalbot: I know what you mean. Any idea how this is done on other open source projects? I could kind of go either way here

stalbot: We could at least be friendly and indicate that almost any valid BQ connection will do if you want to provide your own creds since most (all?) of the tests use public data.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -81,6 +81,8 @@ public static String constructUrlFromPropertiesFile(Properties properties, boole
         dataset = dataset == null ? properties.getProperty("dataset") : dataset;
 
         String forreturn = "";
+        // Set to '?' for the first param and '&' for subsequent params.
+        String paramSep = "?";
```
stalbot, line 85: heh, feels like now that you're in here making changes there's probably a better idiomatic way to build this URL, but in the "touch as few lines as possible" principle this approach wins.

stalbot (approved):
I was thinking this would go in the JDBC url params rather than a props file, just because for our use case it's easier for customers to set the params than the props, but I assume you did that for a good reason based on how you've hooked this into Looker, and if so 👍 . If not, happy to chat through why I think the params might be easier once you get there.

README.md
wnob: I'm just gonna remove this for now. There's probably a better place to put it in g3docs.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
wnob: Unfortunately there doesn't seem to be a standard URL builder for java. I agree something a little more elegant than this would be ideal, but this is pretty simple and easy to understand which is also a plus.

PR #81, opened 2021-03-30
Add support for sending per-statement labels

wnob:
The driver will now accept per-statement labels in addition to the per-connection labels. It checks to make sure it doesn't send more than the [max of 64 labels per query](https://cloud.google.com/bigquery/docs/labels-intro#requirements), but doesn't check anything else about label validity. I have an accompanying PR for Looker to make use of this. This introduces an auxiliary API that is not typical of JDBC drivers and would not normally be used by "typical" users of this driver. As such, it is end-to-end tested in helltool, but not here.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -212,7 +212,7 @@ public class BQConnection implements Connection {
         String userAgent = caseInsensitiveProps.getProperty("useragent");
 
         // extract any labels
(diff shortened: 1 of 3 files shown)


stalbot (commented):
Seems fine, but also seems like it should have a test? Especially if you've got this method you've added that seems like it's for testing.

wnob:
Ya the idea was to use that method in an end-to-end test in helltool. Anyway, I've added a test, which also tests to make sure that the URL labels are correctly parsed and added to the query as well, which is not covered in the helltool test.

stalbot (approved):
👍

PR #83, opened 2021-04-09
Add support for BIGNUMERIC

jswett77:
Allow support for BIGNUMERIC now that it is GA


pom.xml
```diff
@@ -4,7 +4,7 @@
        to a groupId that we have ownership over -->
   <groupId>com.github.jonathanswenson</groupId>
   <artifactId>bqjdbc</artifactId>
-  <version>2.3.4-SNAPSHOT</version>
+  <version>2.3.5</version>
```
jswett77, line 7: No snapshot to release, then when released, create a second PR to tag w/ snapshot correct?

PR #84, opened 2021-04-12
Add support for BIGNUMERIC now that it is GA

jswett77:
Add support for BIGNUMERIC

### DIFF
--- a/src/test/java/net/starschema/clouddb/jdbc/PreparedStatementTests.java
+++ b/src/test/java/net/starschema/clouddb/jdbc/PreparedStatementTests.java
@@ -35,8 +35,7 @@ import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 
-import junit.framework.Assert;
-
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


src/test/java/net/starschema/clouddb/jdbc/PreparedStatementTests.java
```diff
@@ -170,7 +171,8 @@ public void ResultSetMetadataFunctionTestTypes() {
                 java.sql.Types.NUMERIC,
                 java.sql.Types.TIMESTAMP,
                 java.sql.Types.TIME,
-                java.sql.Types.VARCHAR
+                java.sql.Types.VARCHAR,
+                java.sql.Types.NUMERIC
```
stalbot, line 175: I worry this doesn't test `getString` and `getObject` ... did you manually confirm at least that those worked?

jswett77: I ran a `select * from` with a table that had bignumeric columns in sql runner. I can add those tests here though too.

jswett77:
Ok closing this, I merged it to a new branch in this project so it could build in the right context

PR #88, opened 2021-05-26
Record total_bytes_processed and cache_hit on results sets

wnob:
I'm unsure exactly how to assert in the test since these values may or may not be subject to change. Documentation is underspecified. This seems to work.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
@@ -167,11 +167,11 @@ public class BQStatement extends BQStatementRoot implements java.sql.Statement {
                     return new BQForwardOnlyResultSet(
                             this.connection.getBigquery(),
                             projectId,
-                            referencedJob, this, qr.getRows(), fetchedAll, qr.getSchema());
+                            referencedJob, this, qr.getRows(), fetchedAll, qr.getSchema(), qr.getTotalBytesProcessed(), qr.getCacheHit());
                 } else if (fetchedAll) {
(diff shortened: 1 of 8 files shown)


wnob:
(Confirmed with Alexey that bytes_processed will always be zero when cache_hit is true)

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -167,11 +167,11 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
                     return new BQForwardOnlyResultSet(
                             this.connection.getBigquery(),
                             projectId,
-                            referencedJob, this, qr.getRows(), fetchedAll, qr.getSchema());
+                            referencedJob, this, qr.getRows(), fetchedAll, qr.getSchema(), qr.getTotalBytesProcessed(), qr.getCacheHit());
```
stalbot, line 170: I think there are a couple of these below that you that want to also change (line 210 or so)... probably?

stalbot (commented):
Yep this is probably hard to test well... if manual tests worked I'm fine with these tests as they are. 

Just the one comment about a couple extra places that might matter where a result set is init'd

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
wnob: In those cases the data gets populated in the constructors (lines 160-161 of BQForwardOnlyResultSet and lines 85-86 of BQScrollableResultSet). In the forward-only case, it needs to actually fetch the results first, and in the scrollable case, it's contained in the query results object passed to the constructor.

PR #91, opened 2021-06-08
Support geography type

mkou:


wnob (approved):
So this simply treats all `GEOGRAPHY` objects as strings? LGTM

PR #92, opened 2021-07-02
Fix NPE due to unboxing primitive objects

wnob:


kaibankslooker:
Hey @wnob, nice change, I've been messing around with this too. Here's a template for a `BQForwardOnlyResultSetFunctionTest` regression test, if you'd like to add one: 

```Java
@Test
public void testResultSetProcedures() throws SQLException, ParseException {
    final String sql = "CREATE PROCEDURE looker_test.procedure_test_3(target_id INT64)\n"
        + "BEGIN\n"
        + "END;";

    this.NewConnection(false);
    java.sql.ResultSet result = null;
    try {
        Statement stmt = BQForwardOnlyResultSetFunctionTest.con
            .createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
        stmt.setQueryTimeout(500);
        result = stmt.executeQuery(sql);
    } catch (SQLException e) {
        this.logger.error("SQLexception" + e.toString());
        Assert.fail("SQLException" + e.toString());
    }

    System.out.println(result.toString());
}
```

Need to do some cleanup so that we don't repeatedly try to create a routine with the same name, and polish the assertions a bit, but... might be helpful as a starting point.

PR #96, opened 2021-07-12
Fix result set NPE for procedure creation during async query

kaibankslooker:
Super similar to https://github.com/jonathanswenson/bqjdbc/pull/92. 

Found that async queries suffer from the same poor primitive unboxing as sync queries, resulting in NPE.


stalbot (approved):
👍

PR #99, opened 2021-07-14
Add support for 'queryCache' parameter

wnob:
This adds a new query parameter `queryCache` that can be used to disable BigQuery's cache. By default, the cache is enabled, but if the user sets this param to anything other than `true` (case-insensitive), it will be disabled.

Also does a bit of refactoring for the labels logic so that per-statement labels are no longer ignored when the query execution logic goes through `BQStatementRoot` rather than `BQStatement`. I'm not sure this would ever be an issue for Looker, but figured it's worth patching and adding a unit test.


mkou (approved):
LGTM 👍

PR #107, opened 2021-08-11
Add `timeoutMs` URL param

thomasbanghart:
Addresses FR raised [here](https://yaqs.corp.google.com/eng/q/4918674062948958208?team_name=2398870489016565760) by adding support for a `timeoutMs` param which overrides the default query timeout of `Integer.MAX_VALUE / 1000 - 1` 

The FR mentions the [`jobTimeoutMs` param](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#jobconfiguration) but the driver only creates a `JobConfiguration` for prepared statements which (from my understanding) aren't used in Looker. So, I figured `timeoutMs` would get close enough.

Happy to change anything or drop it altogether -- seemed like an easy add 👍

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
@@ -118,6 +118,14 @@ public class BQStatement extends BQStatementRoot implements java.sql.Statement {
     }
 
     @Override
+    public int getQueryTimeout() {
+        if (this.connection.getJobTimeoutMs() != null) {
+            return this.connection.getJobTimeoutMs().intValue();
+        }
+        return this.querytimeout * 1000;
+    }
+
+    @Override
     protected Map<String, String> getAllLabels() {
         return
             ImmutableMap.<String, String>builder()
@@ -262,9 +270,9 @@ public class BQStatement extends BQStatementRoot implements java.sql.Statement {
                 // application bandwidth.
                 Thread.sleep(500);
                 this.logger.debug("slept for 500" + "ms, querytimeout is: "
-                        + this.querytimeout + "s");
+                        + getQueryTimeout() + "ms");
             }
-            while (System.currentTimeMillis() - this.starttime <= (long) this.querytimeout * 1000);
+            while (System.currentTimeMillis() - this.starttime <= (long) getQueryTimeout());
             // it runs for a minimum of 1 time
         } catch (IOException e) {
             throw new BQSQLException(
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
@@ -120,7 +120,7 @@ public class BQStatement extends BQStatementRoot implements java.sql.Statement {
     @Override
     public int getQueryTimeout() {
         if (this.connection.getTimeoutMs() != null) {
-            return this.connection.getTimeoutMs().intValue();
+            return this.connection.getTimeoutMs();
         }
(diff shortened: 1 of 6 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -262,9 +270,9 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
                 // application bandwidth.
                 Thread.sleep(500);
                 this.logger.debug("slept for 500" + "ms, querytimeout is: "
-                        + this.querytimeout + "s");
+                        + getQueryTimeout() + "ms");
             }
-            while (System.currentTimeMillis() - this.starttime <= (long) this.querytimeout * 1000);
+            while (System.currentTimeMillis() - this.starttime <= (long) getQueryTimeout());
```
thomasbanghart, line 275: The `timeoutMs` is normally a query param for the [`getQueryResults` endpoint](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) but since we're polling the jobs table in this loop it made sense to me to change the condition here.

jswett77:
LGTM, @mkou , @wnob - any reservations?

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -117,6 +117,14 @@ public void setLabels(Map<String, String> statementLabels) {
         this.statementLabels = ImmutableMap.copyOf(statementLabels);
     }
 
+    @Override
+    public int getQueryTimeout() {
+        if (this.connection.getTimeoutMs() != null) {
+            return this.connection.getTimeoutMs().intValue();
+        }
+        return this.querytimeout * 1000;
```
mkou, line 125: is `this.querytimeout` always set ?

mkou (approved):
👍  LGTM I just have one question above

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
thomasbanghart: Yes, it's set as a constant in the parent class https://github.com/jonathanswenson/bqjdbc/blob/5ccf45b3f3cbf766a444bd25d83456cf88c047ee/src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java#L71

thomasbanghart:
Thanks for reviewing! I don't have any permissions on the repo so someone else who does may need to do the merge 🙃 
Could either of you make me a contributor if you have a sec?

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -117,6 +117,14 @@ public void setLabels(Map<String, String> statementLabels) {
         this.statementLabels = ImmutableMap.copyOf(statementLabels);
     }
 
+    @Override
+    public int getQueryTimeout() {
+        if (this.connection.getTimeoutMs() != null) {
+            return this.connection.getTimeoutMs().intValue();
```
wnob, line 123: nit: this can be omitted and the `Integer` would be auto-unboxed

thomasbanghart:
If tests fail this time it might be due to the fact I'm trying to merge from a fork and can't access the "secrets" for the tests. I'll close this and open a branch off the main repo if this is the case.

thomasbanghart:
Yep, gonna close this out and reopen as a branch on this repo see if that works -- the secret env variables look to be empty during test set up

PR #108, opened 2021-08-12
Add `timeoutMs` URL param

thomasbanghart:
Addresses FR raised [here](https://yaqs.corp.google.com/eng/q/4918674062948958208?team_name=2398870489016565760) by adding support for a `timeoutMs` param which overrides the default query timeout of `Integer.MAX_VALUE / 1000 - 1` 

The FR mentions the [`jobTimeoutMs` param](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#jobconfiguration) but the driver only creates a `JobConfiguration` for prepared statements which (from my understanding) aren't used in Looker. So, I figured `timeoutMs` would get close enough.

Happy to change anything or drop it altogether -- seemed like an easy add 👍 

Clone of closed PR: https://github.com/jonathanswenson/bqjdbc/pull/107

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -71,6 +71,8 @@ public class BQConnection implements Connection {
 
     private Long maxBillingBytes;
 
+    private Integer timeoutMs;
+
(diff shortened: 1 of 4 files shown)


wnob (approved):
LGTM

PR #109, opened 2021-08-13
Allow totalBytesProcessed and cacheHit to be null

wnob:
Being null can be significant, especially in light of row-level ACLs, for which `totalBytesProcessed` is set to null to prevent users from inferring things they shouldn't know about the table. Let's just let the total bytes processed and cache hit metrics be null. We'll just neglect to emit the Pinger metrics in Looker if they're null.

Also add some more tests. Testing null-readiness is really annoying since so many things can be null. Open to suggestions on how I might test this better.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
@@ -191,15 +191,14 @@ public class BQStatement extends BQStatementRoot implements java.sql.Statement {
                             (qr.getRows() != null && qr.getTotalRows().equals(BigInteger.valueOf(qr.getRows().size()))));
             // Don't look up the job if we have nothing else we need to do
             if (!(fetchedAll || this.connection.isClosed())) {
-                if (qr.getJobReference() == null) {
-                    throw new BQSQLException("Got BQ response with no job reference: " + qr.toString());
+                if (qr.getJobReference() != null) {
+                    referencedJob =
+                        this.connection.getBigquery()
+                            .jobs()
+                            .get(projectId, qr.getJobReference().getJobId())
+                            .setLocation(qr.getJobReference().getLocation())
+                            .execute();
                 }
-                referencedJob =
-                    this.connection.getBigquery()
-                        .jobs()
-                        .get(projectId, qr.getJobReference().getJobId())
-                        .setLocation(qr.getJobReference().getLocation())
-                        .execute();
             }
             if (jobComplete) {
                 if (resultSetType != ResultSet.TYPE_SCROLL_INSENSITIVE) {
--- a/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
@@ -179,7 +179,7 @@ public class BQStatement extends BQStatementRoot implements java.sql.Statement {
         }
 
         this.starttime = System.currentTimeMillis();
-        Job referencedJob;
+        Job referencedJob = null;
         int retries = 0;
         boolean jobAlreadyCompleted = false;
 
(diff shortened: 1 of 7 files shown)


src/test/java/net/starschema/clouddb/jdbc/BQScrollableResultSetFunctionTest.java
```diff
@@ -722,7 +727,48 @@ public void TestResultSetRelative() {
...
+                    new MockLowLevelHttpResponse().setContent(jsonResponse))
+                .build();
+        BQConnection bq = new BQConnection(url + "&useLegacySql=false", properties, mockTransport);
+        BQStatement stmt = new BQStatement(properties.getProperty("projectid"), bq, ResultSet.TYPE_SCROLL_INSENSITIVE, ResultSet.CONCUR_READ_ONLY);
+        String sqlStmt = "SELECT word from publicdata:samples.shakespeare LIMIT 100";
+
+        BQScrollableResultSet results = ((BQScrollableResultSet)stmt.executeQuery(sqlStmt));
+
+        results.getTotalBytesProcessed();
+        results.getCacheHit();
```
jswett77, line 772: So we're just verify that there aren't any NPEs then?

jswett77: Oh nvrmind, missed the opening commend :D

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -190,33 +190,30 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
...
+            if (!(fetchedAll || this.connection.isClosed())) {
+                if (qr.getJobReference() != null) {
+                    referencedJob =
+                        this.connection.getBigquery()
+                            .jobs()
+                            .get(projectId, qr.getJobReference().getJobId())
+                            .setLocation(qr.getJobReference().getLocation())
+                            .execute();
+                }
+            }
```
jswett77, line 202: I'm a little lost w/o context here - what was causing the NPE?

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -190,33 +190,30 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
...
-                    Boolean cacheHit = defaultValueIfNull(qr.getCacheHit(), false);
-                    Long totalBytesProcessed = defaultValueIfNull(qr.getTotalBytesProcessed(), 0L);
                     List<TableRow> rows = defaultValueIfNull(qr.getRows(), new ArrayList<TableRow>());
                     TableSchema schema = defaultValueIfNull(qr.getSchema(), new TableSchema());
 
                     return new BQForwardOnlyResultSet(
                             this.connection.getBigquery(),
                             projectId,
-                            referencedJob, this, rows, fetchedAll, schema, totalBytesProcessed, cacheHit);
+                            referencedJob, this, rows, fetchedAll, schema, qr.getTotalBytesProcessed(), qr.getCacheHit());
```
jswett77, line 211: Or is more about allowing for nullable total bytes processed and cache hit values?

jswett77 (commented):
This feels pretty specific to an error I don't have full context for. I think it looks good, happy to +1 but @kaibankslooker do you have any reservations/suggestions?

src/test/java/net/starschema/clouddb/jdbc/BQScrollableResultSetFunctionTest.java
wnob: Ya, I'm suspicious that defaulting to `false` and 0 for `cache_hit` and `total_bytes_processed`, respectively, is creating some reporting issues. This could help explain why the cache hit rate we're calculating from the BQ job metadata logs (~40%) is so different from the hit rate we're calculating from these pinger metrics (~16%); defaulting to `false` when it should really be "no comment".

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
wnob: In this particular case, it was https://github.com/jonathanswenson/bqjdbc/blame/0f81d63e747b1e1922bea07bd8936198ecf85ec5/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java#L160

But this has been a recurring issue unfortunately :/ It actually turns out that every single getter in the `QueryResponse` object might be null, according to the API, it was just never an issue before because basically the only ones that turn out to be null in practice are `cache_hit` and `total_bytes_processed`.

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
wnob: It's both issues. The more pressing issue if fixing the NPE, but the reporting issue is also a concern.

src/test/java/net/starschema/clouddb/jdbc/BQScrollableResultSetFunctionTest.java
jswett77: Ok. When we integrate this new build of the bq driver, will we be able to test this from HT?

kaibankslooker: ^ happy to help in the ht-side testing effort. Good call.

kaibankslooker (approved):
LGTM - like Justin mentioned, we should take the time to build corresponding ht-side regression tests.

PR #110, opened 2021-08-13
Format with google-java-format

kalenp:
For consistency with other google projects.  My editor kept trying to
enforce this on unrelated changes, so let's do it as a separate pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -206,6 +206,19 @@
           <autoReleaseAfterClose>true</autoReleaseAfterClose>
         </configuration>
       </plugin>
+      <plugin>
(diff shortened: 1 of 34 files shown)


fzakaria:
my 2cents: I'm only inclined to accept if it's automatically enforced and CI fails if it's not.
https://github.com/diffplug/spotless/tree/master/plugin-maven#google-java-format

kalenp:
Updated pom.xml to include the plugin, which also caught a number of files I forgot to reformat.

fzakaria:
I didn't check the pom.xml but please make sure it fails the build.
It should be part of an early lifecycle stage (validate)

Farid Zakaria

PR #111, opened 2021-08-16
Upgrade google clients

kalenp:
Switches from the authentication in
com.google.api.client.googleapis.auth.oauth2.GoogleCredential to
com.google.auth.oauth2.GoogleCredentials.  The former is deprecated and
does not support the newer authentication mechanisms that the latter
does.  This switch should let us use the latest authentication
mechanisms specified in the google credentials json file, such as
external authentication.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,12 +109,17 @@
     <dependency>
       <groupId>com.google.api-client</groupId>
       <artifactId>google-api-client</artifactId>
-      <version>1.25.0</version>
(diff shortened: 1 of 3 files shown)


kalenp:
Currently chained off of the formatting change PR to better highlight these specific changes.  I have run the tests locally and they pass, but I have not yet evaluated how comprehensive they are wrt the various authentication schemes.

fzakaria (approved):

PR #113, opened 2021-08-18
Enforce spotless check during validate

kalenp:
I added spotless to maven previously, but forgot to include it in the `verify` step.  This ensures that it will cause CI to fail if bad changes are checked in.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,10 +215,19 @@
         <groupId>com.diffplug.spotless</groupId>
         <artifactId>spotless-maven-plugin</artifactId>
         <version>2.12.2</version>
+        <executions>
+          <execution>
+            <phase>validate</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
         <configuration>
           <java>
             <googleJavaFormat>
-              <version>1.11.0</version>
+              <!-- Latest version which runs on java8 -->
+              <version>1.7</version>
               <style>GOOGLE</style>
             </googleJavaFormat>
           </java>
--- a/src/main/java/net/starschema/clouddb/jdbc/DMDResultSet.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/DMDResultSet.java
@@ -485,7 +485,7 @@ public class DMDResultSet extends ScrollableResultset<Object> implements java.sq
   public void close() throws SQLException {
     super.close();
     this.Colnames = null;
-  };
+  }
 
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -215,10 +215,19 @@
...
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
         <configuration>
           <java>
             <googleJavaFormat>
-              <version>1.11.0</version>
+              <!-- Latest version which runs on java8 -->
+              <version>1.7.0</version>
```
fzakaria, line 230: huh?

pom.xml
```diff
@@ -215,10 +215,19 @@
...
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
         <configuration>
           <java>
             <googleJavaFormat>
-              <version>1.11.0</version>
+              <!-- Latest version which runs on java8 -->
+              <version>1.7</version>
```
fzakaria, line 230: why a lower version?

pom.xml
```diff
@@ -215,10 +215,19 @@
...
         <artifactId>spotless-maven-plugin</artifactId>
         <version>2.12.2</version>
+        <executions>
+          <execution>
+            <phase>validate</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
```
fzakaria, line 225: Was this from the README ?
I would probably add a comment here or in README how to run spotless manually then to fix it.
(I can't remember i they added it to the error output)

fzakaria (commented):
TY

pom.xml
kalenp: The failing output includes the suggestion to run `mvn spotless:apply`.

pom.xml
kalenp: The latest versions won't run on java8, which is what CI is running with.  Something about a newer version of the java class format.

src/main/java/net/starschema/clouddb/jdbc/DMDResultSet.java
```diff
@@ -485,8 +485,7 @@ public DMDResultSet(Object[] objects, String[] colnames, DMDResultSetType type)
   public void close() throws SQLException {
     super.close();
     this.Colnames = null;
-  }
-  ;
+  };
```
kalenp, line 488: Apparently different versions disagree about what to do about extraneous semicolons after function definitions.  I should just remove that instead of reformatting it...

PR #114, opened 2021-08-20
Return Unauthorized when using invalid OAuth token

kalenp:
Updating the google auth library changed the behavior of handling an
invalid oauth token.  Previously, it would return a failure response as
Unauthorized.  With the change, an interaction in the google libraries
would cause it to throw an IllegalStateException.

This change smooths over that interaction so that we can return an
Unauthorized response as we did before.


kalenp:
Bump, if somebody wouldn't mind giving this a look.

thomasbanghart (approved):
LGTM 👍

thomasbanghart:
There are ht tests that check for `Unauthorized` response for invalid tokens right? This should keep those happy too I would think.

kalenp:
Yes, those failing tests are what sent me back here to make sure we continue to produce the same error.

PR #116, opened 2021-09-07
Adds support for service account impersonation 

thomasbanghart:
Adds a new URL param `targetServiceAccount=<String>` to allow for [service account "impersonation"](https://cloud.google.com/iam/docs/impersonating-service-accounts) e.g. `targetServiceAccount=<service-account>@<project>.iam.gserviceaccount.com` 

Choosing an auth method is still required so, for example, `withApplicationDefaultCredentials=true&targetServiceAccount=<service-account-email>` would use the application default as source and attempt to impersonate the target account.

This should work fine for `withServiceAccount` and `withApplicationDefaultCredentials`. I _think_ this should be fine for `oAuthAccessToken` also but I have yet to verify (should probably do that). 

One thing to note, the [docs](https://cloud.google.com/iam/docs/impersonating-service-accounts#allow-impersonation) mention a few IAM roles that would allow impersonation. For the purposes of this driver, the `Service Account Token Creator` role is the one to use -- it should be granted to the _source_ credential and the IAM Service Account APIs need to be enabled for the project. The target account will need the appropriate BigQuery permissions to access the desired project/dataset/tables.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,7 @@ jobs:
           echo "$PROTECTEDACCOUNT_PROPERTIES" > src/test/resources/protectedaccount.properties
           echo "$SERVICE_ACCOUNT_PROPERTIES" > src/test/resources/serviceaccount.properties
           echo "$OAUTHACCOUNT_PROPERTIES" > src/test/resources/oauthaccount.properties
+          echo "$APPLICATION_DEFAULT_PROPERTIES" > src/test/resources/applicationdefault.properties
           echo "Secrets loaded!"
         shell: bash
         env:
@@ -39,6 +40,7 @@ jobs:
           PROTECTEDACCOUNT_PROPERTIES: ${{ secrets.PROTECTEDACCOUNT_PROPERTIES }}
           SERVICE_ACCOUNT_PROPERTIES: ${{ secrets.SERVICE_ACCOUNT_PROPERTIES }}
           OAUTHACCOUNT_PROPERTIES: ${{ secrets.OAUTHACCOUNT_PROPERTIES }}
+          APPLICATION_DEFAULT_PROPERTIES: ${{ secrets.APPLICATION_DEFAULT_PROPERTIES }}
       - name: Build with Maven
         env: 
           GOOGLE_APPLICATION_CREDENTIALS: src/test/resources/bigquery_credentials_protected.json
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -155,6 +155,7 @@ public class BQConnection implements Connection {
     String userKey = caseInsensitiveProps.getProperty("password");
     String userPath = caseInsensitiveProps.getProperty("path");
 
+    String targetServiceAccount = caseInsensitiveProps.getProperty("targetserviceaccount");
     String oAuthAccessToken = caseInsensitiveProps.getProperty("oauthaccesstoken");
 
     // extract withServiceAccount property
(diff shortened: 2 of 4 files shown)


kalenp:
> No special permissions are necessary for the target account.

The target account will need the relevant BigQuery permissions.  Maybe that's obvious, but I think it's worth being explicit that we'll use the target's BQ permissions.

src/test/java/net/starschema/clouddb/jdbc/JdbcUrlTest.java
```diff
@@ -197,6 +197,49 @@ public void canConnectWithApplicationDefaultCredentials() throws SQLException, I
...
+  @Test
+  public void canImpersonateServiceAccountWithApplicationDefaultAsSource()
+      throws IOException, SQLException {
+    Properties testProps = getProperties("/protectedaccount-json.properties");
+    String url =
+        BQDriver.getURLPrefix()
+            + testProps.getProperty("projectid")
+            + "/"
+            + testProps.getProperty("dataset");
+    String targetServiceAccount = "looker-service-account@looker-test-db.iam.gserviceaccount.com";
```
kalenp, line 209: This should be in one of the properties files rather than embedded directly in the code since it's part of Looker's test environment for this package.  If somebody else wants to test this library, they'll want to use a different service account for their test environment.

src/test/java/net/starschema/clouddb/jdbc/JdbcUrlTest.java
```diff
@@ -197,6 +197,49 @@ public void canConnectWithApplicationDefaultCredentials() throws SQLException, I
...
+  public void impersonatingServiceAccountWithLimitedPermissionsWorksAsExpected()
+      throws IOException, SQLException {
+    Properties testProps = getProperties("/protectedaccount-json.properties");
+    String url =
+        BQDriver.getURLPrefix()
+            + testProps.getProperty("projectid")
+            + "/"
+            + testProps.getProperty("dataset");
+    // dummy service account that has the BigQuery User role but no access to the orders table
+    String targetServiceAccount = "limited-bq-access@looker-test-db.iam.gserviceaccount.com";
```
kalenp, line 227: Ditto.

kalenp (changes requested):
Just the one request to the testing credentials.  

That consolidation refactor from the prior change definitely helped keep this change nice and contained.

kalenp (approved):
LGTM!

thomasbanghart:
Thanks for the review! Will fail tests until we can add that properties file but I asked @jonathanswenson on how we could send up a new one 👍

.github/workflows/build.yml
```diff
@@ -27,6 +27,7 @@ jobs:
           echo "$PROTECTEDACCOUNT_PROPERTIES" > src/test/resources/protectedaccount.properties
           echo "$SERVICE_ACCOUNT_PROPERTIES" > src/test/resources/serviceaccount.properties
           echo "$OAUTHACCOUNT_PROPERTIES" > src/test/resources/oauthaccount.properties
+          echo "$APPLICATION_DEFAULT_PROPERTIES" > src/test/resources/applicationdefault.properties
```
jonathanswenson, line 30: Ok I have added this secret (APPLICATION_DEFAULT_PROPERTIES) to the repository that contains the data that @thomasbanghart sent me.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -155,6 +155,7 @@ public BQConnection(String url, Properties loginProp, HttpTransport httpTranspor
     String userKey = caseInsensitiveProps.getProperty("password");
     String userPath = caseInsensitiveProps.getProperty("path");
 
+    String targetServiceAccount = caseInsensitiveProps.getProperty("targetserviceaccount");
```
jonathanswenson, line 158: should be this targetaccount? or targetserviceaccount? Looking at the properties file I added for a secret -- I think that uses targetaccount instead of targetserviceaccount.

jonathanswenson:
Kicked off a re-build of the actions to see if the updates to the repo secrets will fix the tests.

PR #117, opened 2021-09-08
Set default application credentials path for test runs

thomasbanghart:
Sets the `GOOGLE_APPLICATION_CREDENTIALS` environment variable for the `JdbcUrlTest` class so tests that need it are guaranteed to have it set.


kalenp (approved):
Tested locally and this works for me.

PR #118, opened 2021-09-09
add Bi Engine Statistics and jobId

mkou:


pom.xml
```diff
@@ -118,18 +118,18 @@
     </dependency>
     <dependency>
       <groupId>com.google.http-client</groupId>
-      <artifactId>google-http-client-jackson2</artifactId>
-      <version>1.39.2</version>
+      <artifactId>google-http-client-jackson</artifactId>
+      <version>1.29.2</version>
```
wnob, line 122: Why the version downgrade? Maybe worth a code comment.

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -2967,4 +3001,20 @@ public boolean wasNull() throws SQLException {
...
+
+  public @Nullable String getBiEngineMode() {
+    return biEngineMode;
+  }
+
+  public @Nullable List<BiEngineReason> getBiEngineReasons() {
+    return biEngineReasons;
+  }
+
+  public String getJobId() {
```
wnob, line 3013: @Nullable

pom.xml
mkou: We went from jackson to jackson 2 recently and I realized that when I was running it in helltool (even master) it was generating an error 🤷‍♀️

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -86,6 +86,14 @@
   private final @Nullable Long totalBytesProcessed;
   /** Whether the ResultSet came from BigQuery's cache */
   private final @Nullable Boolean cacheHit;
+  /** Specifies which mode of BI Engine acceleration was performed (if any). */
+  private final String biEngineMode;
```
mkou, line 90: @Nullable

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -86,6 +86,14 @@
...
+  /**
+   * In case of DISABLED or PARTIAL bi_engine_mode, these contain the explanatory reasons as to why
+   * BI Engine could not accelerate. In case the full query was accelerated, this field is not
+   * populated.
+   */
+  private final List<BiEngineReason> biEngineReasons;
```
mkou, line 96: @Nullable

src/main/java/net/starschema/clouddb/jdbc/BQScrollableResultSet.java
```diff
@@ -58,6 +60,17 @@
...
+  /** Specifies which mode of BI Engine acceleration was performed (if any). */
+  private final @Nullable String biEngineMode;
+  /**
+   * In case of DISABLED or PARTIAL bi_engine_mode, these contain the explanatory reasons as to why
+   * BI Engine could not accelerate. In case the full query was accelerated, this field is not
+   * populated.
+   */
+  private final @Nullable List<BiEngineReason> biEngineReasons;
+
+  private final JobReference jobReference;
```
mkou, line 72: @Nullable

src/main/java/net/starschema/clouddb/jdbc/BQScrollableResultSet.java
```diff
@@ -257,4 +279,20 @@ public String getString(int columnIndex) throws SQLException {
...
+
+  public @Nullable String getBiEngineMode() {
+    return biEngineMode;
+  }
+
+  public @Nullable List<BiEngineReason> getBiEngineReasons() {
+    return biEngineReasons;
+  }
+
+  public String getJobId() {
```
mkou, line 291: @Nullable

wnob, line 291: @Nullable

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -193,7 +198,29 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
...
+            JobStatistics2 statistics2 = referencedJob.getStatistics().getQuery();
+            if (statistics2 != null) {
+              biEngineStatistics = statistics2.getBiEngineStatistics();
+            }
+          }
+
+          if (biEngineStatistics != null) {
+            biEngineMode = biEngineStatistics.getBiEngineMode();
+            biEngineReasons = biEngineStatistics.getBiEngineReasons();
+          } else {
```
wnob, line 213: Rather than all these else-clauses, how about initializing them to null above the if statement.

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -183,8 +188,8 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
               && (qr.getTotalRows().equals(BigInteger.ZERO)
                   || (qr.getRows() != null
                       && qr.getTotalRows().equals(BigInteger.valueOf(qr.getRows().size()))));
-      // Don't look up the job if we have nothing else we need to do
-      if (!(fetchedAll || this.connection.isClosed())) {
```
wnob, line 187: When is it necessary to query BI engine statistics after having fetched all results? Should that be a test case?

wnob (changes requested):
Some minor nits, but also seems like incomplete test coverage. Is that coming in a follow-up PR?

pom.xml
kalenp: What kind of an error?  I did that upgrade and I noticed that we needed an accompanying change on the Looker side, but I added that and it seemed to be working.

pom.xml
```diff
@@ -4,7 +4,7 @@
        to a groupId that we have ownership over -->
   <groupId>com.github.jonathanswenson</groupId>
   <artifactId>bqjdbc</artifactId>
-  <version>2.3.14-SNAPSHOT</version>
+  <version>2.3.14</version>
```
kalenp, line 7: Until we're ready to release, I believe you want to keep this as -SNAPSHOT.

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
kalenp: For lists, it's often simpler to have them always initialized, but empty instead of null.  Do you need to distinguish null from an empty list here?

kalenp: If we were in kotlin here, I might also recommend a sealed class for BiEngineMode to capture the combination of the string and optional list of reasons, but I'm not sure of the best way to capture that in java that's not too heavy handed.

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -126,13 +145,27 @@ public BQForwardOnlyResultSet(
...
     logger.debug("Created forward only resultset TYPE_FORWARD_ONLY");
     this.Statementreference = (Statement) bqStatementRoot;
     this.bigquery = bigquery;
     this.completedJob = completedJob;
     this.projectId = projectId;
+    BiEngineStatistics biEngineStatistics = null;
+    if (completedJob != null) {
+      biEngineStatistics = completedJob.getStatistics().getQuery().getBiEngineStatistics();
+    }
+    if (biEngineStatistics != null) {
```
kalenp, line 161: This `if` can be folded into the prior one, since the only way to set `biEngineStatistics` is `if (completedJob != null)`.  Separating them here makes me think that there's another path to this code, so I think it hurts readability.

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
kalenp: If you don't explicitly initialize the members, they will default to null, so you don't need to re-set them to null here (unless there's a way that they could get set and then unset, of course).  Similarly, when declaring `BiEngineStatistics biEngineStatistics = null;` the ` = null` is redundant.

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -193,7 +198,29 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
...
+          BiEngineStatistics biEngineStatistics = null;
+          if (referencedJob != null) {
+            JobStatistics2 statistics2 = referencedJob.getStatistics().getQuery();
+            if (statistics2 != null) {
+              biEngineStatistics = statistics2.getBiEngineStatistics();
+            }
+          }
+
+          if (biEngineStatistics != null) {
+            biEngineMode = biEngineStatistics.getBiEngineMode();
```
kalenp, line 211: Similar comment to the other block, this could get moved into the one path that you actually set `biEngineStatistics` above.

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
```diff
@@ -333,21 +335,40 @@ public ResultSet executeQuery(String querySql, boolean unlimitedBillingBytes)
...
           TableSchema schema = defaultValueIfNull(qr.getSchema(), new TableSchema());
+          BiEngineStatistics biEngineStatistics = null;
+          if (referencedJob != null) {
+            biEngineStatistics = referencedJob.getStatistics().getQuery().getBiEngineStatistics();
+          }
+          String biEngineMode = null;
+          List<BiEngineReason> biEngineReasons = null;
+
+          if (biEngineStatistics != null) {
+            biEngineMode = biEngineStatistics.getBiEngineMode();
```
kalenp, line 357: Same.

kalenp (commented):
Some more java styling and readability suggestions.

pom.xml
mkou: definitely down for it then

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
mkou: Since it is nullable in the BQ API, I kept the same format. It also permits to remove the param in the pinger if it is null in heltool like we do with other data

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
mkou: It was raising an exception saying it was not set 🤷‍♀️

pom.xml
mkou: What are the helltool changes you needed to make ?

kalenp: https://github.com/looker/helltool/pull/64605 (already merged)

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
mkou: "a sealed class for BiEngineMode to capture the combination of the string and optional list of reasons" => I would do the same but since they did not in the BigQuery classes I feel they are not quite sealed in stone yet

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
mkou: > Some minor nits, but also seems like incomplete test coverage. Is that coming in a follow-up PR?
Unfortunately I am very limited in terms of tests as we don't currently have a test DB availiable for JDBC with BI Engine enabled.. I tests thanks to HellTool

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
mkou: Apologies still brushing up my java skills

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
mkou: Actually contrary to cacheHit we only get the BiEngine results in the end. I'll double check

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
mkou: https://cloud.google.com/bigquery/docs/reference/rest/v2/Job

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
mkou: Read it like this: Before we needed the referenced job only when fetchedAll was false because we still had to do some stuff with the job. Now even if everything was indeed fetched, we need the referencedJob in order to get the BiEngine Statistics

pom.xml
```diff
@@ -4,7 +4,7 @@
        to a groupId that we have ownership over -->
   <groupId>com.github.jonathanswenson</groupId>
   <artifactId>bqjdbc</artifactId>
-  <version>2.3.14-SNAPSHOT</version>
+  <version>2.3.14-SNAPSHOT2</version>
```
kalenp, line 7: Extraneous change?

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
kalenp: Bump?

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -193,6 +198,18 @@ private ResultSet executeQueryHelper(String querySql, boolean unlimitedBillingBy
                   .get(projectId, qr.getJobReference().getJobId())
                   .setLocation(qr.getJobReference().getLocation())
                   .execute();
+
+          BiEngineStatistics biEngineStatistics = null;
```
kalenp, line 202: I don't see this being used at this level.  You can move the declaration in to the level where it's actually used.  That should help readability, since the reader will know that they only need to track this to that inner scope, rather than across the full function.

src/main/java/net/starschema/clouddb/jdbc/BQStatementRoot.java
```diff
@@ -333,21 +335,40 @@ public ResultSet executeQuery(String querySql, boolean unlimitedBillingBytes)
...
+              .getBigquery()
+              .jobs()
+              .get(projectId, qr.getJobReference().getJobId())
+              .execute();
+
       if (defaultValueIfNull(qr.getJobComplete(), false)) {
         List<TableRow> rows = defaultValueIfNull(qr.getRows(), new ArrayList<TableRow>());
         if (BigInteger.valueOf(rows.size()).equals(qr.getTotalRows())) {
           TableSchema schema = defaultValueIfNull(qr.getSchema(), new TableSchema());
+          BiEngineStatistics biEngineStatistics = null;
```
kalenp, line 349: Similar to previous comment, you can move this declaration inside the `if (referencedJob != null)` check so that its scope is more clearly defined.

kalenp (commented):
Just some readability suggestions.

kalenp (approved):
LGTM.  Thanks for taking the time to make those readability changes.

mkou:
@kalenp of course ! thanks for this thorough review. I love gaining more and more familiarity with java and good practices.

tjbanghart:
Thanks @mkou!

PR #120, opened 2021-10-28
Adds support for SA delegates with impersonation.

tjbanghart:
Adds support for [Service Account delegation](https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-delegated) for impersonation. 

The `targetServiceAccount` param now accepts a comma separated list of SA emails (in order) to use as a delegation chain to access the "target" account.  The last SA will be used as the "target" account for impersonation.

For example `targetServiceAccount=<delegate-sa-1>,<delegate-sa-2>,<target-ccount>`

If a single SA is provided then it will simply be used as the impersonated account.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -315,13 +315,10 @@ public class BQConnection implements Connection {
    * Return null if {@code string} is null. Otherwise, return an array of delegates iff {@code
    * string} can be parsed as an array when split by {@code delimiter}.
    */
-  private static List<String> parseArrayQueryParam(@Nullable String string, Character delimiter)
-      throws BQSQLException {
-    List<String> delegates = null;
-    if (string != null) {
-      delegates = Arrays.asList(string.split("\\s*" + delimiter + "\\s*"));
-    }
-    return delegates;
+  private static List<String> parseArrayQueryParam(@Nullable String string, Character delimiter) {
+    return string == null
+        ? Collections.emptyList()
+        : Arrays.asList(string.split("\\s*" + delimiter + "\\s*"));
   }
 
   /**
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -310,7 +310,7 @@ public class BQConnection implements Connection {
   }
 
   /**
-   * Return an empty list if {@code string} is null. Otherwise, return an array of delegates iff
+   * Return an empty list if {@code string} is null. Otherwise, return an array of strings iff
    * {@code string} can be parsed as an array when split by {@code delimiter}.
    */
   private static List<String> parseArrayQueryParam(@Nullable String string, Character delimiter) {
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -155,13 +155,14 @@ public class BQConnection implements Connection {
     String userKey = caseInsensitiveProps.getProperty("password");
     String userPath = caseInsensitiveProps.getProperty("path");
 
+    // extract a list of "delegate" service accounts leading to a "target" service account to use
+    // for impersonation.
+    // if only a single service account is provided, then it will be used as the "target"
     List<String> targetServiceAccounts =
         parseArrayQueryParam(caseInsensitiveProps.getProperty("targetserviceaccount"), ',');
-    String oAuthAccessToken = caseInsensitiveProps.getProperty("oauthaccesstoken");
 
-    // extract a list of delegates if provided, should be a list of comma seperated SA emails
-    List<String> delegates =
-        parseArrayQueryParam(caseInsensitiveProps.getProperty("delegates"), ',');
+    // extract OAuth access token
+    String oAuthAccessToken = caseInsensitiveProps.getProperty("oauthaccesstoken");
 
     // extract withServiceAccount property
     boolean serviceAccount =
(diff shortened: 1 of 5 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -304,6 +311,19 @@ private static Integer parseIntQueryParam(String param, @Nullable String paramVa
     return val;
   }
 
+  /**
+   * Return null if {@code string} is null. Otherwise, return an array of delegates iff {@code
+   * string} can be parsed as an array when split by {@code delimiter}.
+   */
+  private static List<String> parseArrayQueryParam(@Nullable String string, Character delimiter)
+      throws BQSQLException {
+    List<String> delegates = null;
```
kalenp, line 320: While this driver code is mixed on usage, there's a best practice of never returning null lists, but instead returning empty lists.  See Effective Java 3rd edition, item 54 "Return empty collections or arrays, not nulls".

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -304,6 +311,19 @@ private static Integer parseIntQueryParam(String param, @Nullable String paramVa
...
+    if (string != null) {
+      delegates = Arrays.asList(string.split("\\s*" + delimiter + "\\s*"));
```
kalenp, line 322: Why the extra `\s` on either side of the delimiter?

kalenp (approved):
Straightforward enough.  

I wonder if we want to expose the distinction between delegates and the target SA, or if we should just take a single list as the jdbc parameters and then break it up for the client.  Makes for a simpler client interface, but then again, clients using this probably already know some about the gcp apis, so maybe that's nothing new or them.

Going to approve in case you want to keep it this way, just throwing that out as another possibility for how you could do this.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -304,6 +311,16 @@ private static Integer parseIntQueryParam(String param, @Nullable String paramVa
     return val;
   }
 
+  /**
+   * Return null if {@code string} is null. Otherwise, return an array of delegates iff {@code
```
kalenp, line 315: Should update the comment to match the new code.

tjbanghart:
Hey @kalenp just seeing these comments, we could certainly use the `targetServiceAccount` and accept a list of SAs. In that case I would think the last SA would simply be the target. Happy to make that change 👍

kalenp:
I could go either way, really.  I'll let you use your judgement on what you think clients will want to use, either a single list with all of the SAs or follow with the gcp libraries and have the target SA with a list of delegates.

tjbanghart:
Thanks for the reviews here Kalen. I think having a single list in `targetServiceAccount` feels better to me. I've updated the PR to reflect that. 

If anyone feels strongly we can revert to `targetServiceAccount` + `delegates` before we cut a release.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -304,6 +309,16 @@ private static Integer parseIntQueryParam(String param, @Nullable String paramVa
...
   }
 
+  /**
+   * Return an empty list if {@code string} is null. Otherwise, return an array of strings iff
+   * {@code string} can be parsed as an array when split by {@code delimiter}.
+   */
+  private static List<String> parseArrayQueryParam(@Nullable String string, Character delimiter) {
+    return string == null
+        ? Collections.emptyList()
+        : Arrays.asList(string.split(delimiter + "\\s*"));
```
kalenp, line 319: What's the trailing `\\s*` for?  Handling a comma at the end of the list or something?

src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
```diff
@@ -103,11 +103,11 @@
       HttpTransport httpTransport,
       String userAgent,
       String rootUrl,
-      String targetServiceAccount,
+      List<String> targetServiceAccounts,
       @Nullable String oauthToken) {
 
-    if (targetServiceAccount != null) {
-      credential = impersonateServiceAccount(credential, targetServiceAccount);
+    if (!targetServiceAccounts.isEmpty()) {
```
kalenp, line 109: On style nit is that you could move this check into `impersonateServiceAccount()`.  If it's empty, then you just return the passed in credential, otherwise you perform the impersonation logic.  The benefit there would be that you don't require as much from the caller of `impersonateServiceAccount()`, they just pass in a list, and the impersonate function deals with it being empty in a consistent way.  If we had multiple places calling the function, that would be more apparent, but with just the single call site, it really is just a nit.

kalenp (approved):
Looks good, just some small questions/comments.  I think from the client code's perspective, having a single list will be easier to manage.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
tjbanghart: Any potential whitespace between SA emails -- alternatively we could trim each element after the split. Or, just trust that we'll get a `targetServiceAccount` sans whitespace

For things like this that might sneak in: `<sa-1>@services.com,  <sa-2>@services.com`

src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
tjbanghart: Sounds good, thanks for catching that. Currently, no checking occurs in `impersonateServiceAccount` which does feel wrong. Nothing to stop a caller from passing in an empty array and blowing things up.

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
kalenp: Ah, that makes sense.  Probably good enough for now.  I think split + strip would be more clear, except that this is java, so it might not actually be that clear, because there's no simple `map` functionality.

PR #121, opened 2021-11-09
Mark assertj as a test dependency

jonathanswenson:
This was added as a standard dependency rather than test dependency, correcting that here.


tjbanghart (approved):
LGTM 👍

tjbanghart (approved):

PR #122, opened 2021-11-09
Prevent setting projectId for sync queries twice

tjbanghart:
Looks like we were setting the `projectId` param to be the `querySql` and then overwriting it by calling `setProjectId`. We set the `querySql` while building the `QueryRequest` so we were just doing this incorrectly. Wasn't causing any problems but probably good to correct it.


kalenp (approved):
LGTM, just looks like somebody fat fingered this before.

PR #125, opened 2021-11-15
Prevent NPE if connection is closed before result set is initialized 

tjbanghart:
Edge case was causing a NPE to be thrown when initializing `BQForwardOnlyResultSet`. The underlying issue seems to be a race condition where a connection is closed after the job has been completed but _before_ the driver has an opportunity to get the query results. Specifically, calling `close()` on a `BQConnection` sets its `bigquery` attribute to `null` so efforts to `getQueryResultsDivided` would fail. 

This just adds an additional check to see if `bigquery` is null and throws a more informative exception.


kalenp (approved):
LGTM. Presumably we're also working on the race condition on the client side as well.

tjbanghart:
Yep hunting that down, although it only occurs during Load Runner test runs where >300 virtual users are running queries simultaneously. Might be hitting a limit client side that is causing some connections to be killed but can't say for sure yet.

PR #127, opened 2022-01-25
Check for empty jobs response when polling for query completion

tjbanghart:
A NPE could be thrown within `BQStatement.runSync` for long running queries. Doesn't happen often but if the BQ project has a reservation quota that's reached while polling for a query job to finish then check for its current state might return an incomplete `Job`. 

We now throw an `IOException` in that case which will at least retry the polling query rather than bubbling up a NPE.


kalenp (approved):
LGTM as far as I can tell.  The test coverage definitely helps.

PR #130, opened 2022-02-25
fix: Updated groupId details for new deploy to looker-open-source

drstrangelooker:


drstrangelooker:
@fzakaria FYI the new secrets need to be set up with the right credentials for the com.google.looker-open-source groupId and the new pgp signing key.

PR #131, opened 2022-05-12
feat: set quota project when using impersonation

tjbanghart:
When using SA impersonation we need to explicitly set a quota project that will be used for billing services. Here we add the `projectId` specified in the JDBC URL as the quota project when constructing the `ImpersonatedCredentials` to use. This effectively makes the `projectId` declared in the URL the billing project that will be used for all BQ jobs.


kppk (approved):
lgtm, thanks! :tada:

tjbanghart:
Hey @drstrangelooker do you know what this `Semantic Pull Request` check is about? I dont recall that being on the old repo

jkaster (approved):
LGTM

PR #135, opened 2022-06-08
Upgrade GSON The package com.google.code.gson:gson before 2.8.9 are v…

mkou:
Upgrade GSON The package com.google.code.gson:gson before 2.8.9 are vulnerable to Deserialization of Untrusted Data via the writeReplace() method in internal classes, which may lead to DoS attacks


tjbanghart (approved):
LGTM 👍

PR #140, opened 2022-06-27
Add Google Drive scope for OAuth2

fzakaria:
* Add Google Drive scope to all locations where we create Google
Credentials.
* Changed the Drive scope to be read-only.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
@@ -287,8 +287,7 @@ public class Oauth2Bigquery {
       throws GeneralSecurityException, IOException {
     GoogleCredentials credential =
         createServiceAccountCredential(
-                serviceaccountemail, keypath, password, jsonAuthContents, false)
-            .createScoped(GenerateScopes(false));
+            serviceaccountemail, keypath, password, jsonAuthContents, false);
 
     logger.debug("Authorized?");
 
@@ -342,7 +341,6 @@ public class Oauth2Bigquery {
       throws IOException {
     GoogleCredentials credential =
         GoogleCredentials.getApplicationDefault().createScoped(GenerateScopes(false));
-    ;
 
     logger.debug("Authorizing with Application Default Credentials.");
 


src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
```diff
@@ -285,7 +287,8 @@ public static Bigquery authorizeViaService(
       throws GeneralSecurityException, IOException {
     GoogleCredentials credential =
         createServiceAccountCredential(
-            serviceaccountemail, keypath, password, jsonAuthContents, false);
+                serviceaccountemail, keypath, password, jsonAuthContents, false)
+            .createScoped(GenerateScopes(false));
```
tjbanghart, line 291: nit: probably don't need to reset scopes here since `createServiceAccountCredential` handles that.

tjbanghart (approved):
LGTM 👍 You can address the single nit but will work just the same. Thanks!

src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
```diff
@@ -86,7 +86,7 @@ public class Oauth2Bigquery {
   /** Application name set on bigquery connection */
   static final String applicationName = "BigQuery JDBC Driver";
 
-  private static final String DRIVE_SCOPE = "https://www.googleapis.com/auth/drive";
+  private static final String DRIVE_SCOPE = "https://www.googleapis.com/auth/drive.readonly";
```
fzakaria, line 89: Assuming this is fair to change for all uses.
I can't imagine why we would need not a readonly scope.

tjbanghart: BQ does support [the `CREATE EXTERNAL TABLE` DDL statement ](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_external_table_statement) so some might want write access to Drive as well.

tjbanghart: That said, I think it's fine to make it readonly for now and change it back should someone request it.

tjbanghart (approved):

PR #153, opened 2023-02-02
fix: NPE while logging job status

tanclary:
BigQuery has a new feature called Query Queues that will queue the queries when the customer reaches the concurrency limit that can be supported by their slot capacity. https://cloud.google.com/bigquery/docs/query-queues

Previously, when a user reaches maximum concurrency, their query would fail with the error: Quota exceeded:: Your project_and_region exceeded quota for concurrent queries. With query queue, the job status will remain in PENDING while it is being queued. When it is in this state, job startTime is not set which causes the NPE.


src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -370,7 +370,7 @@ public static String logAndGetQueryState(Job pollJob) throws IOException {
             + " ; "
             + pollJob.getJobReference().getJobId()
             + " ; "
-            + (System.currentTimeMillis() - pollJob.getStatistics().getStartTime()));
+            + (System.currentTimeMillis() - pollJob.getStatistics().getCreationTime()));
```
justinseinlin, line 373: The original issue came about because we didn't check `getStartTime()` is not null or empty in the check above. 

Although I don't think it is possible for `getCreationTime()` to be null or empty in this case, I think we can go a bit further in this fix by wrapping it in an `Optional`?

Something like:

```java
Optional.ofNullable(pollJob.getStatistics().getCreationTime()).ifPresent(...);
```

tanclary: I added a commit that addresses this. Let me know if you want me to alter it. Thanks!

src/test/java/net/starschema/clouddb/jdbc/CancelTest.java
```diff
@@ -196,6 +196,20 @@ public void handlesNullResultsWhilePollingJob()
...
+    // test verifies that exception is thrown if one or more of these is missing.
+    Job testJob =
+        new Job()
+            .setStatus(new JobStatus().setState("PENDING"))
+            .setStatistics(new JobStatistics().setCreationTime(12345567789L))
+            .setJobReference(new JobReference().setJobId("Job Id"));
+    String result = BQSupportFuncts.logAndGetQueryState(testJob);
+    Assert.assertEquals(result, "PENDING");
+  }
+
```
justinseinlin, line 212: Add a test for the throwing case?  Otherwise LGTM

justinseinlin (approved):

tjbanghart:
Thanks for the PR @tanclary! There seems to be an issue with permissions for the GitHub actions test run. I'm going to close this PR and open one with these changes: https://github.com/looker-open-source/bqjdbc/pull/156

fzakaria:
Why close and re-open?

tjbanghart:
@tanclary does not yet have "collaborator" access to the repo which is required for presubmit tests to run. They have requested access and are awaiting approval but it is unclear how long this will take. I opened a new PR with their commits in the meantime. This requirement is something I think we should revisit because it blocks community based PRs from being verified (another example https://github.com/looker-open-source/bqjdbc/pull/151, which was merged in f00d793ab2a344c430830deed85aede60cbf1993). 

On that note, I think the maintainers of this repo should be given "owner" permissions for the `looker-open-source` org so we can address these kind of repo management issues ourselves. At the very least, we'll need to be able to edit the GitHub [`secrets`](https://github.com/looker-open-source/bqjdbc/blob/f00d793ab2a344c430830deed85aede60cbf1993/.github/workflows/build.yml#L33-L43) periodically as these service accounts are not guaranteed to exist in perpetuity. I am in contact with the Looker Dev Rel team about getting admin/owner permissions.

fzakaria:
I think you can approve on a PR to run the workflows for non collaborators.

fzakaria:
@tanclary i just sent you invite as "Triage" role at least for this project.

fzakaria (changes requested):
Please see @tjbanghart clone PR for all my comments.

tjbanghart:
The CI tests will run but once the lint failure is fixed we'll see NPE for the test credentials because non-collaborators cannot access the GitHub `secrets` used -- @tanclary might be given given collaborator access soon making this a moot point. I still think its worth getting some of the maintainers of this repo admin privileges so we can manage all of this ourselves.

I'll close the duplicate PR (https://github.com/looker-open-source/bqjdbc/pull/156) since we're going to wait for the next release cycle to merge this anyways.

fzakaria:
I have Admin privilege so I can raise Tanner to what's needed (I think)

fzakaria:
@tjbanghart he has an invite to a "Triage" role as "outside collaborator" -- I wonder if that's enough.
Previously he had no role whatsoever. Worth seeing on Monday.

Is adding yourself as a Collaborator just adding yourself to the Google org or it's Looker dependent?
It's self-service to join the Alphabet Github org 
https://screenshot.googleplex.com/AaSUufkwex7WBHD

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -358,19 +356,23 @@ public static String getQueryState(Job myjob, Bigquery bigquery, String projectI
    *     <p>if the request to get the job specified by myjob and projectId fails
    */
   public static String logAndGetQueryState(Job pollJob) throws IOException {
+    JobStatistics stats = pollJob.getStatistics();
```
fzakaria, line 359: technically if pollJob is null this will fail alreaedy.

fzakaria: Should go below.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -358,19 +356,23 @@ public static String getQueryState(Job myjob, Bigquery bigquery, String projectI
...
     if (pollJob == null
         || pollJob.isEmpty()
         || pollJob.getStatus().isEmpty()
-        || pollJob.getStatistics().isEmpty()) {
+        || stats.isEmpty()) {
       throw new IOException("Failed to fetch query state.");
     }
+    long startTime = Stream.of(stats.getStartTime(), stats.getCreationTime())
```
fzakaria, line 366: order matters here.
Do we want creationTime or startTime first?

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -35,13 +35,11 @@
 import java.math.BigInteger;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.UUID;
+import java.util.*;
```
fzakaria, line 38: Keep as non wildcard please.
Change your intelliJ settings.

fzakaria (approved):

fzakaria:
Run 'mvn spotless:apply' to fix these violations.

tjbanghart:
Merged in https://github.com/looker-open-source/bqjdbc/pull/157

PR #155, opened 2023-02-03
fix: NullPointException when statistics query is null

tjbanghart:
Uses [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) to make fetching BI Engine statistics null-safe.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
@@ -26,14 +26,41 @@ import com.google.api.client.json.JsonGenerator;
 import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.client.util.Data;
 import com.google.api.services.bigquery.Bigquery;
-import com.google.api.services.bigquery.model.*;
-import java.io.*;
+import com.google.api.services.bigquery.model.BiEngineReason;
+import com.google.api.services.bigquery.model.BiEngineStatistics;
+import com.google.api.services.bigquery.model.GetQueryResultsResponse;
+import com.google.api.services.bigquery.model.Job;
+import com.google.api.services.bigquery.model.JobStatistics;
+import com.google.api.services.bigquery.model.JobStatistics2;
+import com.google.api.services.bigquery.model.TableFieldSchema;
+import com.google.api.services.bigquery.model.TableRow;
+import com.google.api.services.bigquery.model.TableSchema;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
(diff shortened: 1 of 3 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
```diff
@@ -22,14 +22,7 @@
...
 
-import com.google.api.services.bigquery.model.BiEngineReason;
-import com.google.api.services.bigquery.model.BiEngineStatistics;
-import com.google.api.services.bigquery.model.Job;
-import com.google.api.services.bigquery.model.JobReference;
-import com.google.api.services.bigquery.model.JobStatistics2;
-import com.google.api.services.bigquery.model.QueryResponse;
-import com.google.api.services.bigquery.model.TableRow;
-import com.google.api.services.bigquery.model.TableSchema;
+import com.google.api.services.bigquery.model.*;
```
jswett77, line 25: I think it is frowned upon to use wildcard includes.

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -157,14 +157,15 @@ public BQForwardOnlyResultSet(
...
     this.bigquery = bigquery;
-
-    if (completedJob != null) {
-      BiEngineStatistics biEngineStatistics =
-          completedJob.getStatistics().getQuery().getBiEngineStatistics();
-      if (biEngineStatistics != null) {
-        biEngineMode = biEngineStatistics.getBiEngineMode();
-        biEngineReasons = biEngineStatistics.getBiEngineReasons();
-      }
+    Optional<BiEngineStatistics> biEngineStatisticsOptional =
```
jswett77, line 160: Had to go read how Optional works, think I understand it now... so you're using Optional to chain calls through completedJob via `map`. (So I guess map isn't passing nulls on and you're guaranteed your BiEngineStatistics is present and can extract the remaining properties. Ok, cool.

jswett77 (commented):
Think this makes sense, probably should fix up that include wild card first though.

src/main/java/net/starschema/clouddb/jdbc/BQStatement.java
tjbanghart: Good catch! Fixed.

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
tjbanghart: Props to @bobdxiao for the fix!

jswett77 (approved):
LGTM

PR #156, opened 2023-02-04
fix: Prevent NPE while logging job status

tjbanghart:
Modify how elapsed time is calculated to use creationTime rather than startTime to avoid NPE when logging query state.


src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -364,14 +365,19 @@ public static String logAndGetQueryState(Job pollJob) throws IOException {
...
-    BQSupportFuncts.logger.info(
-        "Job status: "
-            + pollJob.getStatus().getState()
-            + " ; "
-            + pollJob.getJobReference().getJobId()
-            + " ; "
-            + (System.currentTimeMillis() - pollJob.getStatistics().getStartTime()));
-    return pollJob.getStatus().getState();
+
+    if (Optional.ofNullable(pollJob.getStatistics().getCreationTime()).isPresent()) {
```
fzakaria, line 369: nit: just check if it's null if the optional is never used again.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -364,14 +365,19 @@ public static String logAndGetQueryState(Job pollJob) throws IOException {
...
+      BQSupportFuncts.logger.info(
+          "Job status: "
+              + pollJob.getStatus().getState()
+              + " ; "
+              + pollJob.getJobReference().getJobId()
+              + " ; "
+              + (System.currentTimeMillis() - pollJob.getStatistics().getCreationTime()));
```
fzakaria, line 376: do we actually want to log creation time or start time?

fzakaria, line 376: Maybe this instead:
```
Stream.of(startTime, creationTime).filter(Objects::nonNull).findFirst().orElseThrow(() -> "This is really bad")
```
(order depends on above)
CC @tanclary

fzakaria: (which is preferred and why -- please comment)
Is creationTime always set but startTime isn't?

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
```diff
@@ -364,14 +365,19 @@ public static String logAndGetQueryState(Job pollJob) throws IOException {
...
+      return pollJob.getStatus().getState();
+    } else {
+      throw new IOException("Failed to fetch creation time.");
```
fzakaria, line 379: can this ever happen?
If not make it IllegalStateException or just set a default value for the nullable like 1970.

fzakaria (commented):
Few questions.
Can we add information from the e-mail to the description describing the motivation and the BigQuery Queues please.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
fzakaria: On second thought don't even nest inside if ... try to avoid deep nesting
```
long creationTime = Optional.ofNullable(pollJob.getStatistics().getCreationTime()).getOrElse(pollJob.getStatistics().getStartTime())
```

is yet another option.

tjbanghart:
Closing this in favor of https://github.com/looker-open-source/bqjdbc/pull/153

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
tanclary: According to the internal bug: when a query is queued, its statistics do not include 'startTime', but they do include 'creationTime'. So when the logger previously tried to calculate the elapsed time, an NPE was thrown (hence the switch).

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
justinseinlin: We should not have a CreationTime empty; nor should the JobReference be null as well since it would indicate there is a job created in BQ with incomplete information or the `jobs.get` API returned a job with missing info.  Neither situation is desirable.

src/main/java/net/starschema/clouddb/jdbc/BQSupportFuncts.java
justinseinlin: I think this is trying to log the elapsed time of the job, so falling back to creation_time in the absence of start_time makes sense.  As mentioned in another comment, the creation_time should not be empty.

PR #159, opened 2023-02-10
inspect IOException immediate cause if there are no details in the exception

shivanesabharwal:


src/main/java/net/starschema/clouddb/jdbc/BQSQLException.java
```diff
@@ -176,7 +176,10 @@ private static String enrichReason(String reason, Throwable cause) {
         return prefix + googleJsonResponseException.getDetails().getMessage();
       }
     } else {
-      return prefix + cause.toString();
+      String throwableDetails = cause.toString();
+      // if the `cause` itself does not have a detailed message, ask for the message of its cause.
+      if (!throwableDetails.isEmpty()) throwableDetails = cause.getCause().toString();
```
tjbanghart, line 181: Could `getCause` be null here?

shivanesabharwal: indeed, i'm fixing

src/main/java/net/starschema/clouddb/jdbc/BQSQLException.java
```diff
@@ -176,7 +176,13 @@ private static String enrichReason(String reason, Throwable cause) {
         return prefix + googleJsonResponseException.getDetails().getMessage();
       }
     } else {
-      return prefix + cause.toString();
+      String throwableDetails = cause.toString();
+      // if the `cause` itself does not have a detailed message, ask for the message of its cause.
+      if (!throwableDetails.isEmpty()) {
+        Throwable parentCause = cause.getCause();
```
shivanesabharwal, line 182: @tjbanghart we could also attempt to walk the cause chain until we find something non-null

tjbanghart: Let's start with this and revisit if it doesn't help.

PR #168, opened 2023-07-13
Update com.google.guava:guava dependency to v32.0.1-jre

shalaniw:
bqjdbc uses com.google.guava:guava as a transitive dependency. It needs to be bumped from 31.1-jre to 32.0.1-jre per
https://osv.dev/vulnerability/GHSA-7g45-4rm6-3mm3.

Also added the google-cloud-bigquery dependency, so that we don't have to pick out specific libraries from the Google Java client.

Since JacksonFactory is deprecated
https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.json.jackson2,
now using com.google.api.client.json.gson.GsonFactory instead.

Remove com.google.api.services.iamcredentials.v1 and now use Explicit
Credential Loading
https://github.com/googleapis/google-auth-library-java/blob/main/README.md#explicit-credential-loading
to generate an access token.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,7 +124,7 @@
     <dependency>
       <groupId>com.google.apis</groupId>
       <artifactId>google-api-services-iamcredentials</artifactId>
-      <version>v1-rev20211203-1.32.1</version>
+      <version>v1-rev20211203-2.0.0</version>
     </dependency>
     <!-- START TEST DEPENDENCIES -->
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -111,16 +111,6 @@
       <artifactId>google-cloud-bigquery</artifactId>
       <version>2.29.0</version>
     </dependency>
-    <dependency>
-      <groupId>com.google.http-client</groupId>
-      <artifactId>google-http-client-jackson2</artifactId>
-      <version>1.42.0</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.apis</groupId>
-      <artifactId>google-api-services-iamcredentials</artifactId>
-      <version>v1-rev20211203-2.0.0</version>
-    </dependency>
     <!-- START TEST DEPENDENCIES -->
     <dependency>
       <groupId>org.assertj</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -117,11 +117,6 @@
       <version>1.42.0</version>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>32.0.1-jre</version>
-    </dependency>
-    <dependency>
       <groupId>com.google.apis</groupId>
       <artifactId>google-api-services-iamcredentials</artifactId>
       <version>v1-rev20211203-2.0.0</version>
--- a/pom.xml
+++ b/pom.xml
@@ -107,14 +107,9 @@
       <version>2.11.0</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
```diff
@@ -376,24 +372,8 @@ public static String generateAccessToken(
...
-    String name = "projects/-/serviceAccounts/" + serviceaccountemail;
-    GenerateAccessTokenRequest request = new GenerateAccessTokenRequest();
-    request.setScope(Collections.singletonList(BigqueryScopes.CLOUD_PLATFORM));
 
-    IAMCredentials.Projects.ServiceAccounts.GenerateAccessToken generateAccessToken;
-    generateAccessToken =
-        iamCredentials.projects().serviceAccounts().generateAccessToken(name, request);
-    GenerateAccessTokenResponse response = generateAccessToken.execute();
-    return response.getAccessToken();
+    return credential.refreshAccessToken().getTokenValue();
```
tjbanghart, line 376: So is `refreshAccessToken` the equivalent to making a IAMCredentials request? I wonder why we weren't using this before. Do you know if this was added in a more recent version of the credentials lib?

src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
```diff
@@ -67,7 +63,7 @@ public class Oauth2Bigquery {
   static final HttpTransport HTTP_TRANSPORT = new NetHttpTransport();
 
   /** Global instance of the JSON factory. */
-  private static final JsonFactory JSON_FACTORY = new JacksonFactory();
+  private static final JsonFactory JSON_FACTORY = new GsonFactory();
```
tjbanghart, line 66: Nice 😎

pom.xml
```diff
@@ -107,29 +107,9 @@
...
-      <artifactId>google-api-services-bigquery</artifactId>
-      <version>v2-rev20220611-1.32.1</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.apis</groupId>
-      <artifactId>google-api-services-iamcredentials</artifactId>
-      <version>v1-rev20211203-1.32.1</version>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigquery</artifactId>
+      <version>2.29.0</version>
```
tjbanghart, line 112: 🚀 so much nicer than 5 separate dependencies

src/main/java/net/starschema/clouddb/jdbc/Oauth2Bigquery.java
shalaniw: Yeah, it looks like the equivalent to an IAMCredentials request. This is the [source code](https://github.com/googleapis/google-auth-library-java/blob/38bdf60189b44171f5d481fa934f4ece60553653/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L497) for `refreshAccessToken`. The initial implementation was 9 years ago and the last edit was 10 months ago.

tjbanghart: Im wondering about the reference to `BigqueryScopes.CLOUD_PLATFORM` - I don't see the scopes being carried over in the request in `refreshAccessToken`.  This is probably fine but I don't trust the tests we have for OAuth in this repo much.

tjbanghart (approved):

PR #176, opened 2023-11-07
Revert "Remove OAuth Token From Query Params"

shivanesabharwal:
Reverts looker-open-source/bqjdbc#171

-- haven't had a chance to do a full integration test with Helltool, so reverting so other folks can cut new releases in the meantime.


shalaniw (approved):
LGTM

PR #180, opened 2023-11-23
Remove OAuth token from query params in API requests

shivanesabharwal:
- we were redundantly setting the OAuth Token in query parameter via BigQueryRequestUserAgentInitializer#initializeBigqueryRequest.
- The HttpRequestTimeoutInitializer which builds the HttpRequest was already setting it in the header of the request, which is the accepted method.
- remove the call to setOauthToken and associated code which is no longer required.
- in service of http://b/294919784

this PR was reverted because we did not do a full integration test between Helltool and BigQuery OAuth with these changes, and we wanted to unblock other PRs. Now, I'm revisiting this issue


shivanesabharwal:
Looking to merge this again... Reverted because we were cutting a new release before this was tested.

tjbanghart (approved):
LGTM

PR #184, opened 2023-12-13
Support stateless queries.

goomrw:
From
https://cloud.google.com/java/docs/reference/google-cloud-bigquery/2.34.1/com.google.cloud.bigquery.BigQuery

> Stateless queries: query execution without corresponding job metadata

Stateless queries are currently in preview.

This PR introduces support for stateless queries via a new JDBC query parameter. Set `jobcreationmode` to a string that matches one of the standard `JobCreationMode` enum values:

- `JOB_CREATION_MODE_UNSPECIFIED`: Unspecified JobCreationMode, defaults to `JOB_CREATION_REQUIRED`.
- `JOB_CREATION_REQUIRED`: Default. Job creation is always required.
- `JOB_CREATION_OPTIONAL`: Job creation is optional. Returning immediate results is prioritized. BigQuery will automatically determine if a Job needs to be created. The conditions under which BigQuery can decide to not create a Job are subject to change. If Job creation is required, JOB_CREATION_REQUIRED mode should be used, which is the default.

See
https://github.com/googleapis/java-bigquery/blob/v2.34.0/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java#L98-L111

For example, the following create a connection with optional job creation:

    DriverManager.getConnection(
        "jdbc:BQDriver:my-project?jobcreationmode=JOB_CREATION_OPTIONAL",
        driverProperties);

Note that this will cause queries to fail if the provided project does not have stateless queries preview enabled.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
@@ -238,8 +238,17 @@ public class BQConnection implements Connection {
     this.useQueryCache =
         parseBooleanQueryParam(caseInsensitiveProps.getProperty("querycache"), true);
 
-    this.jobCreationMode =
-        parseJobCreationMode(caseInsensitiveProps.getProperty("jobcreationmode"));
+    final String jobCreationModeString = caseInsensitiveProps.getProperty("jobcreationmode");
+    if (jobCreationModeString == null) {
+      jobCreationMode = null;
+    } else {
+      try {
+        jobCreationMode = JobCreationMode.valueOf(jobCreationModeString);
+      } catch (IllegalArgumentException e) {
+        throw new BQSQLException(
+            "could not parse " + jobCreationModeString + " as job creation mode", e);
+      }
+    }
 
     // Create Connection to BigQuery
     if (serviceAccount) {
@@ -351,21 +360,6 @@ public class BQConnection implements Connection {
     return string == null
         ? Collections.emptyList()
         : Arrays.asList(string.split(delimiter + "\\s*"));
-  }
-
-  /**
-   * Return a {@link JobCreationMode} or raise an exception if the string does not match a variant.
-   */
-  private static JobCreationMode parseJobCreationMode(@Nullable String string)
-      throws BQSQLException {
-    if (string == null) {
-      return null;
-    }
-    try {
-      return JobCreationMode.valueOf(string);
-    } catch (IllegalArgumentException e) {
-      throw new BQSQLException("could not parse " + string + " as job creation mode", e);
-    }
   }
 
   /**


src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -72,6 +72,32 @@ public class BQConnection implements Connection {
   /** Boolean to determine whether or not to use legacy sql (default: false) * */
   private final boolean useLegacySql;
 
+  /**
+   * Enum that describes whether to create a job in projects that support stateless queries. Copied
+   * from BigQueryImpl
+   */
+  public static enum JobCreationMode {
```
fzakaria, line 79: Can we add a link here to the API docs or maybe even a permalink to the BigQuery code itself that has this?

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -72,6 +72,32 @@ public class BQConnection implements Connection {
...
+    /** If unspecified JOB_CREATION_REQUIRED is the default. */
+    JOB_CREATION_MODE_UNSPECIFIED,
```
fzakaria, line 81: so weird to assign unspecified and then have it be another thing.
Why not just assign the variable to REQUIRED from the offset?

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
```diff
@@ -322,6 +351,21 @@ private static List<String> parseArrayQueryParam(@Nullable String string, Charac
         : Arrays.asList(string.split(delimiter + "\\s*"));
   }
 
+  /**
+   * Return a {@link JobCreationMode} or raise an exception if the string does not match a variant.
+   */
+  private static JobCreationMode parseJobCreationMode(@Nullable String string)
```
fzakaria, line 357: is this  function needed?
Just calls valueOf

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -105,6 +105,8 @@ public class BQForwardOnlyResultSet implements java.sql.ResultSet {
   private String projectId;
   /** Reference for the Job */
   private @Nullable Job completedJob;
+  /** The BigQuery query ID; set if the query completed without a Job */
+  private final @Nullable String queryId;
```
fzakaria, line 109: is this set even if there was a jobId?

fzakaria (commented):
stopped right before the start of the tests

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
fzakaria: My preference is code; linking to permalink should stay since the BigQuery SDK is a public Github repo?

src/main/java/net/starschema/clouddb/jdbc/BQConnection.java
fzakaria: 1. static method on enum 'fromString' ?
2. Do we really want to raise that exception :)

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
fzakaria: Thank you.

src/test/java/net/starschema/clouddb/jdbc/ConnectionFromResources.java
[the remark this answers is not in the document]
fzakaria: SGTM

fzakaria (approved):
LGTM

More can be improved in subsequent commits and are outside the scope of this change.

PR #185, opened 2023-12-19
Fix `findColumn` and `get*(String label)` methods.

goomrw:
The ResultSet `findColumn` methods (`BQForwardOnlyResultSet.findColumn`, `BQScrollableResultSet.findColumn`, and `BQResultSet.findColumn`) called `BQResultSetMetadata.getCatalogName` to determine the name of each column instead of `BQResultSetMetadata.getColumnLabel`.

This was wrong because `getCatalogName` always returns the BigQuery project ID, not the column name.

This bug in turn broke the get methods that are implemented in terms of `findColumn`, e.g. `getString(String)`.

This change fixes `findColumn` and adds tests for methods that use it with some caveats:

- `BQResultSet` has its findColumn fixed but it's not tested. That's because it's unused, even by its tests, which request a `TYPE_SCROLL_INSENSITIVE` result set and thus get a `BQScrollableResultSet`.

- `BQScrollableResultSet` only accepts dates and times that can be parsed as longs, which disagrees with the more widely used `BQForwardOnlyResultSet`. Its date and time accessors are uncovered.

- `getBigDecimal` is excluded because it calls `BigDecimal.setScale` without a rounding mode, which no longer works in modern Java. A separate change can fix this.

### DIFF
--- a/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
@@ -586,29 +586,10 @@ public class BQForwardOnlyResultSet implements java.sql.ResultSet {
     throw new BQSQLFeatureNotSupportedException("deleteRow()");
   }
 
-  /**
-   *
-   *
-   * <h1>Implementation Details:</h1>
-   *
-   * <br>
-   * Not implemented yet.
-   *
-   * @throws BQSQLException
-   */
+  /** {@inheritDoc} */
   @Override
   public int findColumn(String columnLabel) throws SQLException {
-    if (isClosed()) {
-      throw new BQSQLException("This Resultset is Closed");
-    }
-    final ResultSetMetaData metadata = this.getMetaData();
-    int columns = metadata.getColumnCount();
-    for (int i = 1; i <= columns; i++) {
-      if (metadata.getColumnLabel(i).equals(columnLabel)) {
-        return i;
-      }
-    }
-    throw new BQSQLException("No Such column labeled: " + columnLabel);
+    return CommonResultSet.findColumn(columnLabel, getMetaData());
   }
 
   /**
--- a/src/main/java/net/starschema/clouddb/jdbc/CommonResultSet.java
+++ b/src/main/java/net/starschema/clouddb/jdbc/CommonResultSet.java
@@ -1,0 +1,27 @@
+package net.starschema.clouddb.jdbc;
+
+import com.google.cloud.bigquery.BigQuerySQLException;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+
+/** A static utility class of methods common to all {@link java.sql.ResultSet} implementations. */
+class CommonResultSet {
+
+  /**
+   * A common implementation of {@link java.sql.ResultSet#findColumn(String)}
+   *
+   * @param label the column-to-find's label
+   * @param metadata the metadata from the {@link java.sql.ResultSet}
+   * @return the integer index of the labeled column, if it's found
+   * @throws SQLException if no column with the given label is found
+   */
+  static int findColumn(final String label, final ResultSetMetaData metadata) throws SQLException {
+    int columnCount = metadata.getColumnCount();
+    for (int column = 1; column <= columnCount; column++) {
+      if (metadata.getColumnLabel(column).equals(label)) {
+        return column;
+      }
+    }
+    throw new BigQuerySQLException("No such Column labeled: " + label);
+  }
+}
--- a/src/test/java/net/starschema/clouddb/jdbc/CommonTestsForResultSets.java
+++ b/src/test/java/net/starschema/clouddb/jdbc/CommonTestsForResultSets.java
@@ -22,6 +22,13 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+/**
+ * An abstract class that contains tests common to {@link ResultSet} implementations. Its name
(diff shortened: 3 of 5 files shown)


src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -61,7 +61,12 @@
 import java.sql.Statement;
 import java.sql.Time;
 import java.sql.Timestamp;
-import java.util.*;
+import java.util.ArrayList;
```
tanclary, line 64: Thanks for fixing this

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -593,12 +598,13 @@ public void deleteRow() throws SQLException {
    */
   @Override
```
tanclary, line 599: Should we update the javadoc above?

src/main/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSet.java
```diff
@@ -593,12 +598,13 @@ public void deleteRow() throws SQLException {
...
   public int findColumn(String columnLabel) throws SQLException {
-    if (this.isClosed()) {
+    if (isClosed()) {
       throw new BQSQLException("This Resultset is Closed");
     }
-    int columncount = this.getMetaData().getColumnCount();
-    for (int i = 1; i <= columncount; i++) {
-      if (this.getMetaData().getCatalogName(i).equals(columnLabel)) {
+    final ResultSetMetaData metadata = this.getMetaData();
```
tanclary, line 604: Is it worth extracting this duplicate `for` loop into a helper that can be called by both `BQForwardOnlyResultSet` and `BQResultSet`? Up to you

tanclary: Realize its dup'd some more below, maybe worth extracting

src/test/java/net/starschema/clouddb/jdbc/CommonTestsForResultSets.java
```diff
@@ -0,0 +1,274 @@
+package net.starschema.clouddb.jdbc;
```
tanclary, line 1: nit: Should file name end in `Test` to conform with other test files?

src/test/java/net/starschema/clouddb/jdbc/CommonTestsForResultSets.java
```diff
@@ -0,0 +1,274 @@
...
+        rs -> {
+          final Reader reader = getter.apply(rs);
+          final String contents =
+              new BufferedReader(reader).lines().collect(Collectors.joining(""));
+          Assertions.assertThat(contents).isEqualTo(value);
+        });
+  }
+
+  @Test
+  public void testGetAsciiStreamByLabel() throws SQLException {
```
tanclary, line 110: I tried to make sure these tests are aligned with how other tests in this repo are written but it's riddled with inconsistency 😵

tanclary: The other tests in the repo, not yours

src/test/java/net/starschema/clouddb/jdbc/BQScrollableResultSetFunctionTest.java
```diff
@@ -745,4 +747,31 @@ public void testStatelessQuery() throws SQLException {
...
+  @Ignore(
+      "Contradiction between BQSCrollableResultSet and BQForwardOnlyResultSet "
+          + "dates and times: b/317107706")
+  public void testGetDateByLabel() throws SQLException {}
+
+  @Override
+  @Test
+  @Ignore(
+      "Contradiction between BQSCrollableResultSet and BQForwardOnlyResultSet "
+          + "dates and times: b/317107706")
```
tanclary, line 775: Could you double check whether bug ID is confidential? A brief search says no but just to be safe..

tanclary (commented):
Left some questions and suggestions 👍

src/test/java/net/starschema/clouddb/jdbc/CommonTestsForResultSets.java
tanclary: We can clean this thing up one PR at a time 🫡

src/main/java/net/starschema/clouddb/jdbc/CommonResultSet.java
```diff
@@ -0,0 +1,27 @@
...
+   * @throws SQLException if no column with the given label is found
+   */
+  static int findColumn(final String label, final ResultSetMetaData metadata) throws SQLException {
+    int columnCount = metadata.getColumnCount();
+    for (int column = 1; column <= columnCount; column++) {
+      if (metadata.getColumnLabel(column).equals(label)) {
+        return column;
+      }
+    }
+    throw new BigQuerySQLException("No such Column labeled: " + label);
```
fzakaria, line 25: nit: I would have this return Optional and have the other methods throw the exception.

fzakaria (approved):
Added nit about optional but otherwise solid change,.

tanclary:
If we do much more work in this repo it might be a good idea to adopt some practices like:
- having a way of tracking issues or easily correlating a Github PR to a b/ (ideally bidirectional)
- squashing PR commits down to a single one so the commit history is more readable/helpful.

These are some things Calcite does that I think work to its advantage, what do you think @goomrw @fzakaria ?

PR #186, opened 2023-12-20
Refactor connection management in bqjdbc tests 

goomrw:
Tests in bqjdbc use redundant but slightly different code to manage connections; for example, compare [BQForwardOnlyResultSetFunctionTest.NewConnection](https://github.com/looker-open-source/bqjdbc/blob/94614a98819f2e2e05e9ec68fd05c1e592be799a/src/test/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSetFunctionTest.java#L224-L249) with [BQScrollableResultSetFunctionTest.NewConnection](https://github.com/looker-open-source/bqjdbc/blob/94614a98819f2e2e05e9ec68fd05c1e592be799a/src/test/java/net/starschema/clouddb/jdbc/BQScrollableResultSetFunctionTest.java#L258-L288).

These methods also assign connections to static variables, which makes accessing them annoying and hampers test parallelization.

A previous change introduced a [ConnectionFromResources helper class](https://github.com/looker-open-source/bqjdbc/blob/71bc3a9d978b80d3a3de2f8c54a19ff96c86b295/src/test/java/net/starschema/clouddb/jdbc/ConnectionFromResources.java) that can replace this duplicated connection management code with instance-level connections. This might slow down tests in the short term but should allow parallelization later.

Best reviewed by commit!

### DIFF
--- a/src/test/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSetFunctionTest.java
+++ b/src/test/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSetFunctionTest.java
@@ -67,7 +67,6 @@ public class BQForwardOnlyResultSetFunctionTest extends CommonTestsForResultSets
 
   Logger logger = LoggerFactory.getLogger(BQForwardOnlyResultSetFunctionTest.class);
   private Integer maxRows = null;
-  private String defaultProjectId = null;
 
   private Connection connect(final String extraUrl) throws SQLException, IOException {
     return ConnectionFromResources.connect("installedaccount1.properties", extraUrl);
@@ -76,8 +75,6 @@ public class BQForwardOnlyResultSetFunctionTest extends CommonTestsForResultSets
   @Before
   public void setConnection() throws SQLException, IOException {
     connection = connect("&useLegacySql=true");
-    final BQConnection bqConnection = (BQConnection) connection;
-    this.defaultProjectId = bqConnection.getProjectId();
   }
 
   @Before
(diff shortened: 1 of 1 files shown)


src/test/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSetFunctionTest.java
```diff
@@ -61,35 +60,46 @@
...
   @Before
-  public void setup() throws SQLException, IOException {
-    Properties props =
-        BQSupportFuncts.readFromPropFile(
-            getClass().getResource("/installedaccount.properties").getFile());
-    String url = BQSupportFuncts.constructUrlFromPropertiesFile(props, true, null);
-    url += "&useLegacySql=false";
-    this.defaultProjectId = props.getProperty("projectid");
-    this.defaultConn = new BQConnection(url, new Properties());
+  public void setConnection() throws SQLException, IOException {
```
fzakaria, line 76: I think having two @Before can be confusing since the order isn't guaranteed -- someone might cause dependency order.

src/test/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSetFunctionTest.java
```diff
@@ -61,35 +60,46 @@
...
+    connection = connect("&useLegacySql=true");
+    final BQConnection bqConnection = (BQConnection) connection;
+    this.defaultProjectId = bqConnection.getProjectId();
```
fzakaria, line 79: you only set this in one @Before but the other might assume its set already

src/test/java/net/starschema/clouddb/jdbc/BQForwardOnlyResultSetFunctionTest.java
```diff
@@ -831,20 +798,22 @@ private void mockResponse(String jsonResponse) throws Exception {
...
-    final ResultSet result = stmt.executeQuery(StatelessQuery.exampleQuery());
-    final String[][] rows = BQSupportMethods.GetQueryResult(result);
-    Assertions.assertThat(rows).isEqualTo(StatelessQuery.exampleValues());
-
-    final BQForwardOnlyResultSet bqResultSet = (BQForwardOnlyResultSet) result;
-    Assertions.assertThat(bqResultSet.getJobId()).isNull();
-    Assertions.assertThat(bqResultSet.getQueryId()).contains("!");
+  public void testStatelessQuery() throws SQLException, IOException {
+    try (Connection statelessConnection =
+        connect("&useLegacySql=false&jobcreationmode=JOB_CREATION_OPTIONAL")) {
```
fzakaria, line 803: maybe there should be a ConnectionBuilder that removes all this JDBC Url crap from our own tests ?
Seems like an obvious way to cause bugs.

fzakaria (commented):
I got to review this by commit -- you are right; too large otherwise

fzakaria (approved):

PR #208, opened 2026-06-09
prepare for v3.3.1 release

siyumiG:
This change bumps the minor version of BQJDBC driver.


siyumiG:
The tests are failing due to NullPointerExceptions because GitHub Actions does not pass repository secrets to PRs originating from forks, so the database connection tests cannot run.

mnvsngv (approved):

